PR #2, opened 2016-11-08
Adding Style Guide checks

BrunoBernardino:
These checks don't "fail" validation (on `zapier build` for example), but show up on `zapier validate`.

Also added tests for the `checks`, since we had none.

This is missing a version update, as I'm unsure if we want to make any other changes before that.

### DIFF
--- a/src/checks/trigger-has-unique-ids.js
+++ b/src/checks/trigger-has-unique-ids.js
@@ -25,7 +25,7 @@ const triggerHasUniqueIds = {
     if (doubleId !== undefined) {
       //const repr = _.truncate(JSON.stringify(missingIdResult), 50);
       return [
-        `Got a two or more results with the id of ${doubleId}`
+        `Got two or more results with the id of ${doubleId}`
       ];
     }
     return [];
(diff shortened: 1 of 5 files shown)


bryanhelmig:
One thing I'd prefer to do @BrunoBernardino is move these style checks into our Python code for server side validation that encompass the entire spectrum of apps (Python, V2, and CLI). We could then run them on demand on the backend, at upload time, or at QA time!

bcooksey:
@bryanhelmig We've gone back and forth about how best to do this. Agree that at some point, we need this all server side so that we can ensure it. However, it also seems like a good idea to let devs get fast feedback versus waiting until a `zapier push` or `zapier promote` and it getting blocked.

Maybe we could put the checks server-side, and have a `zapier validate --include-style` that would hit an endpoint dedicated to running the complete set of checks. That way:

1. Devs can do a local validate to check for obvious brokenness in their apps (quick, no network)
1. They can also get the full list of errors on demand vs tied to another command (slower, network call)
1. The checks only live in one place

bcooksey:
Gonna close this for now as we are still discussing where we want the different pieces to live.

PR #6, opened 2016-11-28
No fancy path resolving.

### LINKED ISSUE #1: Deprecate String into appTester function?
I've been using what I think is a much less magical way to test methods:

```javascript
const App = require('../index');
const appTester = zapier.createAppTester(App);

// prefer:
appTester(App.authentication.test);
// instead of:
appTester('authentication.test');

// prefer:
appTester(App.resources.member.list.operation.perform);
// instead of:
appTester('member.list');
```

Thoughts @bcooksey? 

> Moved from https://github.com/zapier/zapier-platform-cli/issues/2.

bryanhelmig:
This forces you to pass in a real function that can be found on the App definition. Will close https://github.com/zapier/zapier-platform-core/issues/1.

### DIFF
--- a/src/tools/memoize.js
+++ b/src/tools/memoize.js
@@ -1,0 +1,165 @@
+'use strict';
+
+ /* eslint-disable */
+
+// Extracted from assets/app/common/memoize.js. Author @shauser.
+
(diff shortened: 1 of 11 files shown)


bryanhelmig:
We should clarify in docs the error message "You must pass in a function/array.", basically instead of `z.dehydrate('some.path')` you should do `z.dehydrate(App.some.path)`.

bryanhelmig:
- [x] update docs
- [x] update example apps
- [x] remove the references in `zapier describe` https://github.com/zapier/zapier-platform-cli/issues/5
- [x] prune down `resolve-method-path.js` and accompanying tests
- [ ] mark a breaking change (do we need to add a `CHANGES.md`?)

src/tools/memoize.js
```diff
@@ -0,0 +1,165 @@
...
+// are arguments `a`, `b`, and `c`, `a` will point to a `WeakMap`
+// with a key of `b`, which will point to a `WeakMap` with a key
+// of `c`, which will point to the result of `fn`.
+var memoize = function memoize(fn) {
+  return function () {
+    for (var _len = arguments.length, args = Array(_len), _key = 0; _key < _len; _key++) {
+      args[_key] = arguments[_key];
+    }
+
+    return getMemoizedValue(args) || memoizeValue(args, fn.apply(undefined, args));
```
bryanhelmig, line 161: @stevenhauser has some fixes to provide here for caching `undefined` - should be able to open source this soon instead of vendoring it.

PR #9, opened 2016-12-30
Apply performList && performGet if applicable. 

bryanhelmig:
I like this but I'm not 100% sold on this - let's chat about this @bcooksey. The alternative is that the backend introspects across resources and triggers instead of having it right there.

### DIFF
--- a/src/tools/schema.js
+++ b/src/tools/schema.js
@@ -130,7 +130,6 @@ const prepareApp = (appRaw) => {
 module.exports = {
   compileApp,
   validateApp,
-  convertResourceDos,
   serializeApp,
   prepareApp
 };
--- a/src/tools/schema.js
+++ b/src/tools/schema.js
@@ -20,7 +20,7 @@ const convertResourceDos = (appRaw) => {
       trigger.operation.type = 'hook';
       trigger.operation.outputFields = trigger.operation.outputFields || resource.outputFields;
       if (resource.list && resource.list.operation && resource.list.operation.perform) {
-        trigger.operation.performList = trigger.operation.performList || resource.list.perform;
+        trigger.operation.performList = trigger.operation.performList || resource.list.operation.perform;
       }
       triggers[trigger.key] = trigger;
     }
--- a/src/tools/schema.js
+++ b/src/tools/schema.js
@@ -19,6 +19,9 @@ const convertResourceDos = (appRaw) => {
       trigger.operation.resource = resource.key;
       trigger.operation.type = 'hook';
       trigger.operation.outputFields = trigger.operation.outputFields || resource.outputFields;
+      if (resource.list && resource.list.operation && resource.list.perform) {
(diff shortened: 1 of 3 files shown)


BrunoBernardino (approved):
I think this looks good though it might not seem intuitive for us at a glance, because of the new "resources" concept.

bcooksey:
Ok, this should be good for final review. As I was working on the `performList` and `performGet` copying from resources, I realized that we were doing it inconsistently for other fields that the schema claims we should be, so those are fixed.

And yes, I slipped some data copying tweaks in that were really out of scope for this PR, but they were small and I didn't want to forget.

src/tools/schema.js
```diff
@@ -46,23 +41,59 @@ const convertResourceDos = (appRaw) => {
...
+/* When a trigger/search/create (action) links to a resource, we walk up to
+ * the resource and copy missing properties from resource to the action.
+ */
+const copyPropertiesFromResource = (type, action, appRaw) => {
+  if (appRaw.resources && action.operation && appRaw.resources[action.operation.resource]) {
+    const copyableProperties = ['outputFields', 'sample'];
+    const resource = appRaw.resources[action.operation.resource];
+
+    if (type === 'trigger' && action.operation.type === 'hook') {
+      if (resource.list && resource.list.operation && resource.list.operation.perform) {
```
BrunoBernardino: Instead of this long line we can use `lodash` and `if (_.get(resource, 'list.operation.perform')) {` which will return undefined if any of the properties in that "path" don't exist.

Same for similar code below.

src/tools/schema.js
```diff
@@ -46,23 +41,59 @@ const convertResourceDos = (appRaw) => {
...
-  return dataTools.deepCopy({ triggers, searches, creates });
+  return { triggers, searches, creates };
+};
+
+/* When a trigger/search/create (action) links to a resource, we walk up to
+ * the resource and copy missing properties from resource to the action.
+ */
+const copyPropertiesFromResource = (type, action, appRaw) => {
+  if (appRaw.resources && action.operation && appRaw.resources[action.operation.resource]) {
+    const copyableProperties = ['outputFields', 'sample'];
```
BrunoBernardino, line 58: I thought the correct name to be `copiableProperties`? I _think_ it's a typo, which I'd love to see fixed, even if it's not important 😬

BrunoBernardino (approved):
Love it. It looks great! Just a comment to improve readability of some code, a typo, and that's it!

src/tools/schema.js
bcooksey: I think this is a US vs. UK spelling. I'm finding `copyable` in some online dictionaries, but not the Oxford English Dictionary.

BrunoBernardino: Wow, interesting! Looks so strange to me 😄

PR #11, opened 2017-02-20
Add Search Connector to field schema

BrunoBernardino:
Related to https://github.com/zapier/zapier-platform-schema/pull/9. Tests should re-run and then merge after that gets merged and published to NPM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "zapier-platform-core",
-  "version": "0.10.0",
+  "version": "0.10.1",
   "description": "The core SDK for CLI apps in the Zapier Developer Platform.",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "zapier-platform-core",
-  "version": "0.9.12",
-  "description": "The core SDK for the Zapier platform.",
+  "version": "0.10.1",
```
bryanhelmig: Careful, when running `npm version patch` (like we should be doing) it will increment this - which will push to `0.10.2`. Looks like `0.9.12` missed the bump to `0.10.0`.

bryanhelmig (approved):

xavdid (approved):

PR #15, opened 2017-03-04
stop lowercasing headers, fixes #13

### LINKED ISSUE #13: Headers are being lowercased
Setting a header in a `z.request` results in the header being lowercased when sent out on the wire.

Example code snippet to reproduce:

```javascript
  const promise = z.request(`https://api.lyft.com/oauth/token`, {
    method: 'POST',
    json: {
      code: bundle.inputData.code,
      grant_type: 'authorization_code'
    },
    headers: {
      'Content-Type': 'application/json',
      'Authorization': auth_header
    }
  });
```

Resulted in this request:

```
POST https://api.lyft.com/oauth/token
user-agent: Zapier
content-type: application/json; charset=utf-8
cache-control: no-cache
authorization: <auth_header>
```

xavdid:
Pretty simple fix!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "standard.enable": false
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "standard.enable": false
+}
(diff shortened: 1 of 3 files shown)


xavdid:
For reference, is this the kind of fix we can just merge in or should we stick to always doing a PR just for record's sake?

bcooksey (approved):
Glad this was straightforward

bcooksey:
As for your question about documenting changes, I think the paper trail of a PR nice, particularly when it is a behavior change. For something like nuking a `console.log`, it's reasonable to skip the PR.

PR #17, opened 2017-03-16
Fixes issue where RefreshAuthError would never be thrown

benedikt:


bcooksey:
Thanks for giving this a shot @benedikt. I had actually started working on this earlier today but hadn't had a chance to get my PR up.

This was one of the bugs, but there was some more subtlety happening that #18 addresses, so I think we'll go with that one eventually.

Appreciate the help and sorry for the duplicated effort.

PR #18, opened 2017-03-17
Fix autofresh flag for OAuth2

bcooksey:
For #16

### DIFF
--- a/src/tools/create-app-request-client.js
+++ b/src/tools/create-app-request-client.js
@@ -15,6 +15,7 @@ const addBasicAuthHeader = require('../http-middlewares/before/add-basic-auth-he
 // after middles
 const prepareResponse = require('../http-middlewares/after/prepare-response');
 const logResponse = require('../http-middlewares/after/log-response');
+const throwForStaleAuth = require('../http-middlewares/after/throw-for-stale-auth');
 
 const createAppRequestClient = (input, options) => {
   input = ensurePath(input, '_zapier.app');
@@ -43,7 +44,15 @@ const createAppRequestClient = (input, options) => {
   const httpAfters = [
     prepareResponse,
     logResponse
-  ].concat(ensureArray(app.afterResponse));
+  ];
+
+  if (app.authentication) {
+    if (app.authentication.type === 'oauth2' && _.get(app, 'authentication.oauth2Config.autoRefresh', false)) {
+      httpAfters.push(throwForStaleAuth);
+    }
+  }
+
+  httpAfters.concat(ensureArray(app.afterResponse));
 
   return createRequestClient(httpBefores, httpAfters, options);
 };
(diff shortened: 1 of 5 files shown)


bcooksey:
One important thing to note is that, as implemented now, session auth lost the automatic refresh on a 401. I think this is OK because:

1. The behavior between a perform of `{}` and of `func` was different. The shorthand got the extra handling, the functions did not.
1. I'm less confident that session auth will always use a 401 response as those are all custom solutions. OAuth tends to be more standard cause the API is already implementing a spec
1. We _have_ to get that throw of a refresh out of the generic handler because it is the incorrect behavior for other auth types like basic. The error would result in Zapier throwing a "Cannot call refresh on this type of app" and the dev would be stuck with no way to disable that middleware

Sanity check @bryanhelmig @xavdid?

src/tools/create-app-request-client.js
```diff
@@ -43,7 +44,15 @@ const createAppRequestClient = (input, options) => {
   const httpAfters = [
     prepareResponse,
     logResponse
-  ].concat(ensureArray(app.afterResponse));
+  ];
+
+  if (app.authentication) {
+    if (app.authentication.type === 'oauth2' && _.get(app, 'authentication.oauth2Config.autoRefresh', false)) {
```
xavdid: You don't need to pass `fasle` to `_.get` as missing values return `undefined` by default.

test/userapp/index.js
```diff
@@ -398,7 +441,12 @@ const addRequestHeader = (request, z, bundle) => {
   could be useful for APIs that always return 200 even on errors.
  */
 const changeStatusOnErrorResponses = (response) => {
-  const isJsonResponse = response.getHeader('Content-Type').match(/^application\/json/);
+  const contentType = response.getHeader('Content-Type');
+  if (!contentType) {
+    return response;
+  }
+
+  const isJsonResponse = contentType.match(/^application\/json/);
```
xavdid, line 449: Probably don't need a regex here. Right now, it'll match `application/jsonwhatever`, when you really just want to verify that it's `=== 'application/json`

xavdid (approved):
A couple of minor notes, but it makes sense. 

I super support having consistent behavior across request structures. I like being able to have a predictable result across options. But yeah, this makes sense!

test/userapp/index.js
bryanhelmig: `application/json; charset=utf-8` isn't uncommon - maybe `indexOf() === 0` or something.

bryanhelmig (approved):
I think it is much nicer in a standalone "after" middleware.

test/userapp/index.js
bcooksey: Decided to keep what we had since 1) it is test-only code so not too worried of any performance impacts 2) It reads about the same once you have `indexOf('application/json') === 0`

Issue #14, opened 2017-03-01
Calls to `z.stashFile` throw an error when testing locally


bcooksey:
When trying out hydration with files during local development, calls to `z.stashFile` throw this error:

```
1) PDF Should Get a PDF file:
     request to http://localhost:8000/platform/rpc/cli failed, reason: connect ECONNREFUSED 127.0.0.1:8000
What happened:
  Executing hydrators.getPDF with bundle
  request to http://localhost:8000/platform/rpc/cli failed, reason: connect ECONNREFUSED 127.0.0.1:8000
  FetchError: request to http://localhost:8000/platform/rpc/cli failed, reason: connect ECONNREFUSED 127.0.0.1:8000
      at ClientRequest.<anonymous> (/home/ubuntu/workspace/my-app/node_modules/zapier-platform-core/node_modules/node-fetch/index.js:127:11)
      at emitOne (events.js:77:13)
      at ClientRequest.emit (events.js:169:7)
      at Socket.socketErrorListener (_http_client.js:269:9)
      at emitOne (events.js:77:13)
      at Socket.emit (events.js:169:7)
      at emitErrorNT (net.js:1269:8)
      at nextTickCallbackWith2Args (node.js:458:9)
      at process._tickDomainCallback (node.js:413:17)
```

We need to improve the testing process in some way. Some ideas:

1. Could raise a better error that tells dev this does not work locally
2. Could attempt to stash the file by calling the production rpc
3. Could try to provide a short-circuited version that stashing the files on local disk, maybe in a `.stashed_files` directory.

PR #19, opened 2017-03-17
Enable devs to write local tests against production RPC API

bcooksey:
For https://github.com/zapier/zapier-platform-core/issues/14. This is taking the second option of the listed approaches and opening up the production RPC for local testing.

Long term, the local stashing may be interesting, but this will do in the interim.


bryanhelmig (approved):
Looks good to me - thanks for chasing this down @bcooksey.

xavdid (approved):

PR #22, opened 2017-03-21
Support streamed non-ascii files in z.stashFile

BrunoBernardino:
Solves [this problem](https://zapier-platform.slack.com/archives/C2S5DGF6G/p1489610122403555).

### DIFF
--- a/src/tools/create-file-stasher.js
+++ b/src/tools/create-file-stasher.js
@@ -79,25 +79,37 @@ const createFileStasher = (input) => {
       .then(result => {
         if (isPromise(bufferStringStream)) {
           return bufferStringStream.then((maybeResponse) => {
-            let newBufferStringStream;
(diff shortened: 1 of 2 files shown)


bcooksey (approved):
Fixes the issues we had. Chatting with @BrunoBernardino, we realized that the `.buffer()` is ok because we the request to S3 isn't setup to really stream the data anyway, so no matter what we are going to end up with the entire file in memory at some point.

Could be nice to rework that so we don't have to grab it all at once, but for now, this will work.

bryanhelmig:
Huh, I really thought we had stream support here. IIRC the `FormData` lib could easily handle streams ALA:

```javascript
form.append('my_file', fs.createReadStream('/foo/bar.jpg'));
```

We could write to `/tmp` if it doesn't like in memory streams or something - but it is a bit fishy.

PR #23, opened 2017-03-23
Attempt to fix header collision issue

### LINKED ISSUE #13: Headers are being lowercased

BrunoBernardino:
Fixes #13

### DIFF
--- a/src/tools/request-merge.js
+++ b/src/tools/request-merge.js
@@ -12,9 +12,9 @@ const caseInsensitiveMerge = (requestOne, requestTwo, requestThree) => {
   requestThree = requestClean(requestThree);
 
   // This is a very quick & efficient merge for all of request's properties
-  const mergedRequest = _.merge.apply(_, [requestOne, requestTwo, requestThree]);
+  const mergedRequest = _.merge(requestOne, requestTwo, requestThree);
 
-  // Now to cleanup headers, we start on the last request (the one to keep) and work backwards to add the keys
+  // Now to cleanup headers, we start on the last request (the one with priority) and work backwards to add the keys that don't already exist
   // NOTE: This is done "manually" instead of a _.merge or Object.assign() because we need case-insensitivity
   const mergedRequestHeaders = requestThree.headers || {};
   const requestTwoHeaders = requestTwo.headers || {};
--- a/src/tools/request-merge.js
+++ b/src/tools/request-merge.js
@@ -4,6 +4,29 @@ const _ = require('lodash');
 
 const requestClean = require('./request-clean');
 
+// Remove any matching keys (case-insensitive) from object, in order
+const removeExistingSimilarKeys = (object, keys) => {
+  keys.reverse(); // Last added is first to keep.
+
+  const indexLimit = keys.length - 1;
+  let i = 0;
+
+  while (i <= indexLimit && keys[i]) {
+    const checkingKey = keys[i];
+
+    const foundKeyIndex = _.findIndex(keys, (key) => key !== checkingKey && key.toLowerCase() === checkingKey.toLowerCase());
+
+    if (foundKeyIndex !== -1) {
+      const foundKey = keys[foundKeyIndex];
+      delete object[foundKey];
+      keys.splice(foundKeyIndex, 1);
+      i = -1; // Don't want to miss a thing (cue music)
+    }
+
+    i += 1;
+  }
+};
+
 // Stack requests on top of each other - deeply merging them.
 const requestMerge = (requestOne, requestTwo) => {
   const baseRequest = {
@@ -13,16 +36,26 @@ const requestMerge = (requestOne, requestTwo) => {
       'user-agent': 'Zapier'
     }
   };
+
   const requests = [baseRequest, requestOne, requestTwo];
+
   const request = _.merge.apply(_, requests.map(requestClean));
+
   requests.headers = requests.headers || {};
-  request.headers = Object.keys(request.headers).reduce((coll, key) => {
+  const keys = Object.keys(request.headers);
+
+  request.headers = keys.reduce((coll, key) => {
     let val = request.headers[key];
+
     if (val === requestClean.DROP_DIRECTIVE) {
       delete coll[key];
     }
+
     return coll;
   }, request.headers);
+
+  removeExistingSimilarKeys(request.headers, keys);
+
   return request;
 };
 
--- a/src/tools/request-merge.js
+++ b/src/tools/request-merge.js
@@ -4,27 +4,49 @@ const _ = require('lodash');
 
 const requestClean = require('./request-clean');
 
-// Remove any matching keys (case-insensitive) from object, in order
-const removeExistingSimilarKeys = (object, keys) => {
-  keys.reverse(); // Last added is first to keep.
+// Do a merge with case-insensitive keys, and drop empty header keys
+const caseInsensitiveMerge = (requestOne, requestTwo, requestThree) => {
+  // This creates copies/clones
+  requestOne = requestClean(requestOne);
+  requestTwo = requestClean(requestTwo);
+  requestThree = requestClean(requestThree);
 
-  const indexLimit = keys.length - 1;
-  let i = 0;
+  // This is a very quick & efficient merge for all of request's properties
+  const mergedRequest = _.merge.apply(_, [requestOne, requestTwo, requestThree]);
 
-  while (i <= indexLimit && keys[i]) {
-    const checkingKey = keys[i];
+  // Now to cleanup headers, we start on the last request (the one to keep) and work backwards to add the keys
+  // NOTE: This is done "manually" instead of a _.merge or Object.assign() because we need case-insensitivity
+  const mergedRequestHeaders = requestThree.headers || {};
+  const requestTwoHeaders = requestTwo.headers || {};
+  const requestOneHeaders = requestOne.headers || {};
 
-    const foundKeyIndex = _.findIndex(keys, (key) => key !== checkingKey && key.toLowerCase() === checkingKey.toLowerCase());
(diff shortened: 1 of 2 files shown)


bcooksey (commented):
One hesitation with this approach is that the docs for [Object.keys()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) say

> in the same order as that provided by a for...in loop

And then [for...in loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) says

> iterates over the enumerable properties of an object, in arbitrary order

Can we count on the order provided by `.keys()` to consistently reflect last in? @charettes suggestion about a new object type and storing lower-cased keys but including the casing alongside the value seemed solid, though definitely more work. Thoughts?

BrunoBernardino:
Good point, and indeed, by JS object definition we can't get a guaranteed order on an object's properties/keys, that's why I'm using the same array (so the order is the same in both merging and checking for "duplicates").

That being said, I don't expect this to cause a problem (we're already addressing an edge case that isn't common), but if it _ever_ does, we should probably invest in such a full-fledged solution (that won't be easy or quick at all).

xavdid:
I don't think a full custom object is worth it, but I think we're on the right track here. Instead of adding everything and then removing what looks like duplicates, we do the check when we're modifying the headers. Only add the default key if a case insensitive match doesn't exist already. 

Hopefully we only add defaults in one spot :D

xavdid:
So I've been thinking about this today. 

I think the most important thing is that whatever the user puts in, they get back out. If the API they're using has a case-sensitive header and they enter it correctly only to have us swap the case on them, that's gonna be super frustrating. To that end, it would be most cautious to always check before adding a header via a function like this:

```js
// headers comes in as an object
function addHeader(headers, newKey, newVal) {
  let addKey = true;
  for (let k of Object.keys(headers)) {
    if (k.toLowerCase() === newKey.toLowerCase()) {
      addKey = false;
      break;
    }
  }
  if (addKey) {
    header[newKey] = newVal;
  }
  return headers;
}
```
As long as we use that to add a header anywhere we edit them, we'll be safe. 

That being said, as long as the PR as written definitely won't overwrite user-entered info (and I don't think it will, but i'm a little hazy on when & why `mergeRequests` gets used), then I'm fine with it as is. Liked @BrunoBernardino mentioned, it's a super corner case, we're probably worrying over nothing. 😀

xavdid (commented):
So I've been thinking about this today. 

I think the most important thing is that whatever the user puts in, they get back out. If the API they're using has a case-sensitive header and they enter it correctly only to have us swap the case on them, that's gonna be super frustrating. To that end, it would be most cautious to always check before adding a header via a function like this:

```js
// headers comes in as an object
function maybeAddHeader(headers, newKey, newVal) {
  let addKey = true;
  for (let k of Object.keys(headers)) {
    if (k.toLowerCase() === newKey.toLowerCase()) {
      addKey = false;
      break;
    }
  }
  if (addKey) {
    header[newKey] = newVal;
  }
  return headers;
}
```
As long as we use that to add a header anywhere we edit them, we'll be safe. 

That being said, as long as the PR as written definitely won't overwrite user-entered info (and I don't think it will, but i'm a little hazy on when & why `mergeRequests` gets used), then I'm fine with it as is. Liked @BrunoBernardino mentioned, it's a super corner case, we're probably worrying over nothing. 😀

BrunoBernardino:
I agree and love the suggestion @xavdid — like you mentioned, with this PR as is, we should be covered. Do note that to apply your suggestion would make it so that in a merge, we keep the old headers instead of the new (maybe that's a better behavior? Not sure), and we'd have to change how the requests get merged, so we check for each header before being merged, rather than merging and then checking if there's any duplicates (case-insensitive-wise). I think that'd result in more iterations over the keys — I might be looking at it wrongly, though.

I also think this PR as is keeps the surprises to a minimum, meaning that if I add `{'api-key': 'ab'}` and then `{'Api-Key': 'cd'}`, I'll get `{'Api-Key': 'cd'}`, but with your suggestion I think I'd get `{'api-key': 'ab'}`. Maybe that's fine, but it's a different behavior than what we currently have (same key gets latest value), and we would need to change for any merging, not just duplicate-finding.

BrunoBernardino:
So, are we good to merge @xavdid / @bcooksey ? Or should we add more test cases?

BrunoBernardino:
Alright, after some brainstorming I'll try to do a better merge and loop through the keys less, trying to guarantee the order of the keys, per request, are kept, potentially applying the clean and deleting of keys while doing our own-flavoured merge.

src/tools/request-merge.js
```diff
@@ -4,6 +4,51 @@ const _ = require('lodash');
...
 
+// Do a merge with case-insensitive keys, and drop empty header keys
+const caseInsensitiveMerge = (requestOne, requestTwo, requestThree) => {
+  // This creates copies/clones
+  requestOne = requestClean(requestOne);
+  requestTwo = requestClean(requestTwo);
+  requestThree = requestClean(requestThree);
+
+  // This is a very quick & efficient merge for all of request's properties
+  const mergedRequest = _.merge.apply(_, [requestOne, requestTwo, requestThree]);
```
ibolmo: You can do: `_.merge(requestOne, requestTwo, requestThree)`

src/tools/request-merge.js
```diff
@@ -4,6 +4,51 @@ const _ = require('lodash');
...
+  const requestTwoHeaders = requestTwo.headers || {};
+  const requestOneHeaders = requestOne.headers || {};
+
+  // Check, in order, which keys to add (if they're not duplicate)
+  [requestTwoHeaders, requestOneHeaders].forEach((requestHeaders) => {
+    const existingKeys = Object.keys(mergedRequestHeaders);
+    const requestKeys = Object.keys(requestHeaders);
+
+    // We will loop through every header, and if we find no (case-insensitive) match, we'll add it to mergedRequest
+    requestKeys.forEach((checkingKey) => {
```
ibolmo: Could use a lowerCase hash and just check if one is already found? E.g.

```js
// warning pseudo
seen = {'authorization': ....}
lower = key.toLowerCase();
seen[lower] = seen[lower] || requestHeaders[key];
```

src/tools/request-merge.js
```diff
@@ -4,6 +4,51 @@ const _ = require('lodash');
...
+      // Only add if the key doesn't exist yet
+      if (foundKeyIndex === -1) {
+        mergedRequestHeaders[checkingKey] = requestHeaders[checkingKey];
+      }
+    });
+  });
+
+  // Remove any keys with DROP_DIRECTIVE, after all merging happened
+  Object.keys(mergedRequestHeaders).forEach((key) => {
+    if (mergedRequestHeaders[key] === requestClean.DROP_DIRECTIVE) {
```
ibolmo: I wonder if this could be checked inline while doing the looping above.

ibolmo (commented):
Didn't mean to start a review. Now that I have a bit more time, wanted to help :D

src/tools/request-merge.js
BrunoBernardino: Great question! I tried, and the problem is the `DROP_DIRECTIVE` can be in last request's headers, which means that if I delete it then, the other requests can add it back. We have to wait until all the merging is finished.

src/tools/request-merge.js
BrunoBernardino: Great point. Do you think it'll improve readability or performance considerably? I'm thinking the header object will never have _many_ keys, but I might be missing something.

ibolmo: Up to the team. Personally the hash lookup is very clear.

src/tools/request-merge.js
```diff
@@ -4,6 +4,51 @@ const _ = require('lodash');
...
+  // This creates copies/clones
+  requestOne = requestClean(requestOne);
+  requestTwo = requestClean(requestTwo);
+  requestThree = requestClean(requestThree);
+
+  // This is a very quick & efficient merge for all of request's properties
+  const mergedRequest = _.merge(requestOne, requestTwo, requestThree);
+
+  // Now to cleanup headers, we start on the last request (the one with priority) and work backwards to add the keys that don't already exist
+  // NOTE: This is done "manually" instead of a _.merge or Object.assign() because we need case-insensitivity
```
bcooksey, line 18: This "why" type of note is great. Some of the other comments in this function are more "what" type comments that should be ok to nuke.

src/tools/request-merge.js
bcooksey: @ibolmo See what you are saying. The downside there is that you then loose the original casing, so you'd have to track `seen` plus an object that has the casing we want, or else track the desired key along with the value in the insensitive version. At that point, both seem about equal.

bcooksey (approved):

bryanhelmig:
I'm not 100% up to speed on our current plan of attack but alternate technique - now that Lambda supports Node 6.10 we could use `Proxy` and a `set` trap to provide a nativish interface there.

Also, we have prior art (which we use via `fetch` we could lean on) https://github.com/bitinn/node-fetch/blob/master/src/headers.js.

Issue #4, opened 2016-11-08
Better censoring of ad-hoc sensitive values.


bryanhelmig:
For example - if an endpoint returns `{"access_token": "whatever"}` we're not "intelligently" interpreting the JSON and censoring it - we only censor things sent already in `bundle.authData`.

For example - Sentry does intelligent scrubbing in their clients for variable names/configs/etc. ALA https://github.com/getsentry/sentry/blob/8.10.0/src/sentry/constants.py#L166-L175.

lvh-zapier:
Similarly (not sure if this warrants a separate ticket), short strings are not censored: https://github.com/zapier/zapier-platform-core/blob/c06fb327155663d5f2cb6d44562112b18cc2845a/src/tools/create-logger.js#L69-L76

(I'm not sure I understand why, but I haven't really tried to understand the code around it.)

bryanhelmig:
Mostly that is to knock down spamming logs (sometimes a field like `datacenter: a` will result in replacing tons of random `a` strings with needlessly censored values. We're sort of assuming anything shorter than 5 chars is not a proper "secret".

lvh-zapier:
Makes sense. That was my assumption; I'm glad there's no deeper reason I'm missing :)

BrunoBernardino:
The related PR added a test to confirm everything from `bundle.authData` is being censored properly (even inside strings, not just "plain" objects), but the logic of this Feature Request will need to be handled in a separate PR, later (more info in #24).

ronniebermejo:
Wondering if there is any update on censoring ad-hoc response data? thks

BrunoBernardino:
Hi @ronniebermejo, thanks for reaching out!

It's still in the backlog, we're currently prioritizing other more requested features.

We will definitely prioritize this one higher if more people ask for it. Is there a specific set of fields you were looking to auto-censor?

wking:
On Tue, Apr 03, 2018 at 08:50:26AM +0000, Bruno Bernardino wrote:
> We will definitely prioritize this one higher if more people ask for
> it.

Is the best way to do that +1-ing the initial post here?  I've done
that already ;).  @ronniebermejo, on the other hand, went with a bump
comment.  Would it be easier if we consolidated around one approach?
Maybe topic +1s with an additional bump comment if it's been six
months since the last activity?

Also, are there any pointers to public repositories where folks can
chip in on this?  Or is the code for this internal to Zapier?

ronniebermejo:
[...]

PR #24, opened 2017-04-03
Adding test to confirm behavior

BrunoBernardino:
I was adding this test to confirm the behavior for fixing #4 and apparently it's already working?

Or am I misinterpreting the feature request @bryanhelmig ?

### DIFF
--- a/src/tools/create-logger.js
+++ b/src/tools/create-logger.js
@@ -68,7 +68,7 @@ const makeSensitiveBank = (event) => {
   );
   return _.values(sensitiveData)
     .reduce((bank, val) => {
-      // keeps short values from spamming censor strings in logs, <5 chars is not a proper secret
+      // keeps short values from spamming censor strings in logs, < 6 chars is not a proper secret
       // see https://github.com/zapier/zapier-platform-core/issues/4#issuecomment-277855071
       if (val && String(val).length > 5) {
         bank[val] = hashing.snipify(val);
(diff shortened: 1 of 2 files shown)


bryanhelmig:
Ah yeah - this addresses @lvh-zapier's concerns - but I was thinking of the case when we don't censor values that are unknown before to us. For example - during the oauth authentication leg - we only know the `CLIENT_SECRET` - so we can censor that. But the response contains `access_token` which isn't part of our "censor bank" yet.

It would be smart to have logic that perhaps recursed the JSON responses or querystrings and automatically censored stuff with keys like `access_token`, `api_key`, etc. ([Sentry does something like this already in Python](https://github.com/getsentry/sentry/blob/8.10.0/src/sentry/constants.py#L166-L175)).

lvh-zapier:
Huh, I'm confused. I thought this code censored anything of a given size by default already?

lvh-zapier:
(Reminder: I should show y'all some redaction code we have internally, it may have some properties you care about.)

bryanhelmig:
Yeah - I don't think this PR really addresses the core concern of that issue @lvh-zapier - it latched onto an off the cuff comment instead as I wasn't super clear.

BrunoBernardino:
Seems that will require quite a few more things, so I'll merge this in for now and keep the issue open. Thanks for the clarification @bryanhelmig !

PR #26, opened 2017-04-11
Only allow z.stashFile in hydrators

### LINKED ISSUE #3: Block `z.stashFile` in polling endpoint.
This would be very expensive if done in a loop - for example.

Could either:

1. Limit it to N usages per trigger - definitely more permissive.
2. Block all usage not in a hydrate function.

BrunoBernardino:
Fixes #3

### DIFF
--- a/src/tools/create-file-stasher.js
+++ b/src/tools/create-file-stasher.js
@@ -75,6 +75,11 @@ const createFileStasher = (input) => {
       return ZapierPromise.reject(new Error('rpc is not available'));
     }
 
+    const isRunningOnHydrator = _.get(input, '_zapier.event.method', '').indexOf('hydrators.') === 0;
+    if (!isRunningOnHydrator) {
+      return ZapierPromise.reject(new Error('Cannot stash files outside an hydration function/method.'));
+    }
+
     return rpc('get_presigned_upload_post_data')
       .then(result => {
         if (isPromise(bufferStringStream)) {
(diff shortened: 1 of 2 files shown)


xavdid (approved):

src/tools/create-file-stasher.js
```diff
@@ -75,6 +75,11 @@ const createFileStasher = (input) => {
       return ZapierPromise.reject(new Error('rpc is not available'));
     }
 
+    const isRunningOnHydrator = _.get(input, '_zapier.event.method', '').indexOf('hydrators.') === 0;
+    if (!isRunningOnHydrator) {
+      return ZapierPromise.reject(new Error('Cannot stash files outside an hydration function/method.'));
```
bcooksey, line 80: Is there anything we need to do so that tests trying to check a hydrate method don't break? My read of this is that tests that call a hydration method directly would need to define the `_zapier.event.method`? Is that right, or will that get auto generated during testing? If not, wonder if we could make the test harness smart enough to add that in.

BrunoBernardino: On my tests I've been using `appTester(App.hydrators.methodName)` to test file stashing/hydration, and that gets the `_zapier.event.method` filled as expected. Running the function directly will not work, unless they pass in a mock `z`.

I think this is fine because I don't expect developers to test the platform core in their apps directly, right?

xavdid:
[CLI - Block z.stashFile in polling](https://trello.com/c/IQvWDpqt/97-cli-block-z-stashfile-in-polling)

src/tools/create-file-stasher.js
bcooksey: Yup, as long as we provide a way to get that added automatically.

PR #28, opened 2017-04-17
Upgrade node-fetch

BrunoBernardino:
Fixes issues with some variants of `multipart/form-data` uploads.

More info in https://github.com/bitinn/node-fetch/blob/master/CHANGELOG.md#v163


ibolmo (approved):

xavdid (approved):
there were a ton of changes, but from what I read I think this is good to go

PR #30, opened 2017-04-18
Updating searchOrCreate automatic generation for resources

BrunoBernardino:


bcooksey (approved):
If the schema is valid, then we know that we can use the searchOrCreate key and the search interchangeably, so in theory we shouldn't have to swap this. However, being explicit about the current limitation can also be good. Can see it either way.

bcooksey:
Whoops, read this wrong, good find @BrunoBernardino!

PR #33, opened 2017-05-12
Initial stab at locking in core

BrunoBernardino:
This still requires the backend changes that haven't been made, and it's related to [this RFC](https://github.com/zapier/zapier/issues/11866).

Thoughts @bryanhelmig / @bcooksey ?

### DIFF
--- a/src/tools/create-locker.js
+++ b/src/tools/create-locker.js
@@ -1,0 +1,54 @@
+'use strict';
+
+const _ = require('lodash');
+
+const LOCK_MAX_TIMEOUT = 60 * 60 * 24; // 1 day, in seconds
+
+// Designed to be some user provided function/api.
+const createLocker = (input) => {
+  const rpc = _.get(input, '_zapier.rpc');
+
+  return (callback, options) => {
+    if (!rpc) {
+      throw new Error('rpc is not available');
+    }
+
+    if (!options) {
+      options = {};
+    }
+
+    const defaultOptions = {
+      timeout: 60,
+      name: 'single-lock'
+    };
+
+    _.assign(options, defaultOptions);
+
+    if (!options.name) {
+      throw new Error('Lock name needs to be defined.');
+    }
+
+    if (options.timeout > LOCK_MAX_TIMEOUT) {
+      throw new Error(`Timeout for lock is too big. Max is ${LOCK_MAX_TIMEOUT}.`);
+    }
+
+    const isRunningOnTrigger = _.get(input, '_zapier.event.method', '').indexOf('triggers.') === 0;
+    const isRunningOnSearch = _.get(input, '_zapier.event.method', '').indexOf('searches.') === 0;
+    const isRunningOnHydrator = _.get(input, '_zapier.event.method', '').indexOf('hydrators.') === 0;
+    if (isRunningOnSearch || isRunningOnTrigger || isRunningOnHydrator) {
+      throw new Error('Cannot lock inside a trigger, search, or hydration function/method.');
+    }
+
+    return rpc('run_with_lock', options)
+      .then((result) => {
+        if (result.available) {
+          callback();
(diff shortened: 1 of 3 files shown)


src/tools/create-locker.js
```diff
@@ -0,0 +1,54 @@
...
+    if (isRunningOnSearch || isRunningOnTrigger || isRunningOnHydrator) {
+      throw new Error('Cannot lock inside a trigger, search, or hydration function/method.');
+    }
+
+    return rpc('run_with_lock', options)
+      .then((result) => {
+        if (result.available) {
+          callback();
+        } else {
+          // TODO: Do we throw something else here for the runner to catch it?
```
BrunoBernardino: I'm curious about your thoughts for this part mainly.

src/tools/create-locker.js
```diff
@@ -0,0 +1,57 @@
...
+    if (isRunningOnSearch || isRunningOnTrigger || isRunningOnHydrator) {
+      return ZapierPromise.reject(new Error('Cannot lock inside a trigger, search, or hydration function/method.'));
+    }
+
+    return rpc('run_with_lock', options)
+      .then((result) => {
+        if (result.available) {
+          return true;
+        } else {
+          // TODO: Do we throw something else here for the runner to catch it?
```
BrunoBernardino: I'm curious about your thoughts for this part mainly.

bcooksey: So callers are either chaining off the returned promise or doing `await`, correct?

If so, are we:

1. At risk of Lambda timeouts because the process spins while we wait for the lock?
2. Able to retry just the code that needs to lock? Or would a rejection mean eventually retrying the entire caller?

BrunoBernardino: Correct.

1. Yes, but I don't think we should "wait" for more than just the RPC call. It's similar to stashing a file, right now, for example.
2. No. I tried that pattern, but it seemed brittle. If we just send some code to run, we won't be able to reference preexisting variables, etc. I think we'll need to run everything, basically a lock is a "starting point" for when the code would either just run once/synchronously (task-wise) or stop.

src/tools/create-locker.js
```diff
@@ -0,0 +1,57 @@
...
+      options = {};
+    }
+
+    const defaultOptions = {
+      timeout: 60,
+      name: 'single-lock'
+    };
+
+    // We're nice, so we allow a lock without settings
+    _.assign(options, defaultOptions);
```
bryanhelmig: We ought to use and mutate a copy of `options`.

src/tools/create-locker.js
```diff
@@ -0,0 +1,57 @@
...
+
+    if (!options.name) {
+      return ZapierPromise.reject(new Error('Lock name needs to be defined.'));
+    }
+
+    if (options.timeout > LOCK_MAX_TIMEOUT) {
+      return ZapierPromise.reject(new Error(`Timeout for lock is too big. Max is ${LOCK_MAX_TIMEOUT}.`));
```
bryanhelmig: Let's make sure we match these max settings on our backend side.

src/tools/create-locker.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    const isRunningOnTrigger = _.get(input, '_zapier.event.method', '').indexOf('triggers.') === 0;
```
bryanhelmig: I think you might need to cover the resources cases?

src/tools/create-locker.js
bryanhelmig: We could add a `waitFor` but it shouldn't be greater than 20 seconds or so.

I think just raising the error is fine in the default case (maybe `waitFor === 0`).

BrunoBernardino:
Closing this over another approach we're taking.

PR #34, opened 2017-05-19
Add checks for searches

xavdid:
Ensure that they return an array (of any length). Relates to https://github.com/zapier/zapier/issues/11924. 

Once merged, I'll do a `patch` release (this shouldn't break existing code, because before today searches that returned objects would have seen an error on the server) and bump cli.

### DIFF
--- a/test/checks.js
+++ b/test/checks.js
@@ -2,14 +2,19 @@
 
 require('should');
 
-const createIsSingle = require('../src/checks/create-is-single');
+const checks = require('../src/checks');
 
 describe('checks', () => {
   it('createIsSingle sees create return values with multiples', () => {
-    createIsSingle.run('some.method', []).length.should.eql(0);
-    createIsSingle.run('some.method', [{}]).length.should.eql(0);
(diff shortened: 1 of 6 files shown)


test/checks.js
```diff
@@ -2,14 +2,19 @@
 
 require('should');
 
-const createIsSingle = require('../src/checks/create-is-single');
+const checks = require('../src/checks');
 
 describe('checks', () => {
```
BrunoBernardino: If we could add some tests for the other checks (even if simple), it'd be 💯

BrunoBernardino (approved):
This looks great, David! I've got a wishlist item, but it's not required 😄

xavdid:
@BrunoBernardino added tests for all checks! Actually found a bug, so it was probably worth the time. 😁 

I'll merge and deploy on monday!

BrunoBernardino:
Awesome stuff!

bcooksey (approved):
Solid test coverage, and nice refactor of importing on the checks

PR #41, opened 2017-07-10
Throw error if middleware does not return an object

FokkeZB:
Resolves https://github.com/zapier/zapier/issues/12574


BrunoBernardino (approved):
This looks great!

PR #43, opened 2017-07-17
Improving ID checks

### LINKED ISSUE #25: ID Check needs to be defensive
The [trigger-has-id check](https://github.com/zapier/zapier-platform-core/blob/cbd56fa1d54d6fec5927438364a4de8821b33159/src/checks/trigger-has-id.js#L15) needs to be careful about doing `.id` on undefined and other bad values. If an request returns an array of nulls or the like, this raises a `TypeError` instead of the intended message.

### LINKED ISSUE #32: I think we're requiring hook payloads to contain an `id`
AFAICT we're doing the trigger has id check on hook trigger data. Is this intentional? Makes sense for polling triggers, but seems odd that we're also running it against hooks.

https://github.com/zapier/zapier-platform-core/blob/master/src/checks/trigger-has-id.js

BrunoBernardino:
Remove check of .id property in Hooks and make it more defensive (don't throw an exception when the `result` is `null`, for example).

Fixes #25 
Fixes #32

### DIFF
--- a/src/app-middlewares/after/checks.js
+++ b/src/app-middlewares/after/checks.js
@@ -25,7 +25,7 @@ const checkOutput = (output) => {
   if (runChecks) {
     const rawResults = checks
       .filter((check) => {
-        return check.shouldRun(event.method);
+        return check.shouldRun(event.method, event.bundle);
       })
       .map((check) => {
         return check.run(event.method, output.results)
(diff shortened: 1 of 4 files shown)


xavdid (approved):
Looks great! I now realize I meant to grab these and totally spaced it.

PR #44, opened 2017-07-31
Throwing a better error when create returns a list with one item

BrunoBernardino:
Complementary to https://github.com/zapier/zapier/pull/12555

### DIFF
--- a/src/checks/create-is-single.js
+++ b/src/checks/create-is-single.js
@@ -11,7 +11,7 @@ const createIsSingle = {
   name: 'createIsSingle',
   shouldRun: isCreate,
   run: (method, results) => {
-    if (_.isArray(results) && results.length > 1) {
+    if (_.isArray(results)) {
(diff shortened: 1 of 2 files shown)


src/checks/create-is-single.js
```diff
@@ -11,7 +11,7 @@ const createIsSingle = {
   name: 'createIsSingle',
   shouldRun: isCreate,
   run: (method, results) => {
```
kola-er, line 13: Just curious, what's the purpose of argument `method`. It seems not used.

kola-er (approved):
Good to go. Would love to hear your response to my curiosity left in the comment.

src/checks/create-is-single.js
BrunoBernardino: It isn't used here, but it's used in other checks, like this one: https://github.com/zapier/zapier-platform-core/blob/master/src/checks/is-search.js#L3

PR #45, opened 2017-08-28
Allow file stashing within a create/action

benzapier:
Needed for situations where an action/create receives the file's contents from the API/endpoint, and needs to stash that for use in future Zap steps.

### DIFF
--- a/src/tools/create-file-stasher.js
+++ b/src/tools/create-file-stasher.js
@@ -75,11 +75,10 @@ const createFileStasher = (input) => {
       return ZapierPromise.reject(new Error('rpc is not available'));
     }
 
-    const isRunningOnHydratorOrCreate = () => {
-      const whereAreWe = _.get(input, '_zapier.event.method', '');
-      return whereAreWe.indexOf('hydrators.') === 0 || whereAreWe.indexOf('creates.') === 0;
-    };
-    if (!isRunningOnHydratorOrCreate) {
+    const isRunningOnHydrator = _.get(input, '_zapier.event.method', '').indexOf('hydrators.') === 0;
+    const isRunningOnCreate = _.get(input, '_zapier.event.method', '').indexOf('creates.') === 0;
+
+    if (!isRunningOnHydrator && !isRunningOnCreate) {
       return ZapierPromise.reject(new Error('Files can only be stashed within a create or hydration function/method.'));
     }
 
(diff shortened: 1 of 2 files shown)


test/tools/file-stasher.js
```diff
@@ -110,7 +110,7 @@ describe('file upload', () => {
       .catch(done);
   });
 
-  it('should fail if not being called from an hydrator event', (done) => {
+  it('should fail if being called from a trigger event', (done) => {
```
BrunoBernardino, line 113: Looks good! Please consider adding a test for a search (and maybe passing ones for hydration and create) since you're being specific here.

src/tools/create-file-stasher.js
```diff
@@ -75,9 +75,12 @@ const createFileStasher = (input) => {
       return ZapierPromise.reject(new Error('rpc is not available'));
     }
 
-    const isRunningOnHydrator = _.get(input, '_zapier.event.method', '').indexOf('hydrators.') === 0;
-    if (!isRunningOnHydrator) {
-      return ZapierPromise.reject(new Error('Cannot stash files outside an hydration function/method.'));
+    const isRunningOnHydratorOrCreate = () => {
```
BrunoBernardino: This looks good! I'd suggest considering keeping `isRunningOnHydrator` and adding a `isRunningOnCreate`, then you can do the check with `if (!isRunningOnHydrator && !isRunningOnCreate)`. I think it might become more clear and easier to add a `isRunningOnSearch` later if we find it necessary. Again, no need to take the suggestion!

BrunoBernardino (changes requested):
This is looking great! I'd only like to see an extra test since we're specifically mentioning the trigger now, ideally test search, hydration, and create as well (proper passing and failing conditions).

src/tools/create-file-stasher.js
benzapier: Great idea - done!

test/tools/file-stasher.js
benzapier: All set!

benzapier:
@BrunoBernardino Can you double check things, and see if it looks all-clear for merging?

BrunoBernardino (approved):
This looks really clear and well-tested now! Great job! 👍

BrunoBernardino:
Oh, and please don't release a new version after merging this. I'm expecting to ship other things this week and will take care of releasing a new version with these changes included this week or next.

PR #46, opened 2017-09-14
Increase dehydrator max payload size to 2 KB

kola-er:
Had to increase the dehydrator max payload size to accommodate Zap cases that didn't experience 800 bytes size limit on old Python apps.

### DIFF
--- a/test/hydration.js
+++ b/test/hydration.js
@@ -45,6 +45,17 @@ describe('hydration', () => {
       result.should.eql('hydrate|||{"type":"method","method":"some.path.to","bundle":{}}|||hydrate');
     });
 
+    it('should not accept payload size bigger than 2048 bytes.', () => {
+      const inputData = { key: 'a'.repeat(2049) };
+      const payloadSize = JSON.stringify(inputData).length;
+      try {
+        dehydrate(funcToFind, inputData);
+        '1'.should.eql('2'); // shouldn't pass
+      } catch (err) {
+        err.message.should.containEql(`Oops! You passed too much data (${payloadSize} bytes) to your dehydration function - try slimming it down under 2048 bytes (usually by just passing the needed IDs).`);
+      }
+    });
+
   });
 
 });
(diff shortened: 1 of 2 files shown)


test/hydration.js
```diff
@@ -45,6 +45,16 @@ describe('hydration', () => {
       result.should.eql('hydrate|||{"type":"method","method":"some.path.to","bundle":{}}|||hydrate');
     });
 
+    it('should not accept payload size bigger than 4096 bytes.', () => {
+      const inputData = { key: 'a'.repeat(4097) };
+      const payloadSize = JSON.stringify(inputData).length;
+      try {
+        dehydrate(funcToFind, inputData);
```
BrunoBernardino: I'd add a `should(false).eql(true)` to make sure this test fails if `dehydrate` allows a payload with `4097`.

BrunoBernardino (changes requested):
Looks good and simple. I only have a request to improve the test and then it should be good to go.

BrunoBernardino:
@bryanhelmig any concerns on raising this, since you added this limit initially?

To clarify, in this `dehydrate` situation, a very long string (an absolute file `path`) was necessary (id wouldn't work) to pass into the method.

test/hydration.js
kola-er: I've always wondered the purpose of `'1'.should.eql('2'); // shouldn't pass`. Would I be correct if I say `createDehydrator` isn't stable?

BrunoBernardino: It should be stable. That kind of assertion just makes sure that if someone changes the behavior of the tested method, then the test starts failing, otherwise, if the `try` doesn't throw an exception, the test will pass, not doing its job. Does that make sense?

BrunoBernardino (approved):

test/hydration.js
kola-er: Makes perfect sense. Thanks for the explanation.

bryanhelmig:
So the reason we wanted dehydrate to stay small was the state (values encoded into the dehydrate string) gets put into a URL as it is passed around as a pointer to other apps, and since URLs have a finite length, this can break things!

bryanhelmig:
FWIW, I believe it is 4096 for our server configs, so we should be fine with 2k (I think this leaves room for HTTP headers too).

BrunoBernardino:
Thanks for the explanation @bryanhelmig !

PR #47, opened 2017-09-14
PT3 - Allowing any content-type when stashing files

BrunoBernardino:
Related to https://github.com/zapier/zapier/pull/13286 and needs it deployed first, too.


bryanhelmig (approved):
Looks good! For a second I thought we might have been missing the default but I was just getting twisted in how the ordering of the logic is! 😂

PR #48, opened 2017-09-18
Allows modules to be objects in app definitions

ibolmo:
I ran into an edge case where I wanted to define various portions of the app definition in modules.

E.g. authentication:

```js
// ./authentication.js
export const test = () => {};
export const oauth2Config = {};
oauth2Config.authorizeUrl = {...};
```

in index.js

```js
import * as authentication from './authentication';

export default {
     authentication,
};
```

This works out of the box, but testing was breaking due to the `isPlainObject` check. I've added a plain module with an empty index.js to grab the constructor used by the runtime. I'll add a test, but wanted to run this by the team first.

### DIFF
--- a/src/tools/data.js
+++ b/src/tools/data.js
@@ -2,9 +2,10 @@
 
 const _ = require('lodash');
 const memoize = require('./memoize');
+const plainModule = require('./plain');
 
 const isPlainObj = (o) => {
-  return o && typeof o == 'object' && o.constructor === Object;
+  return o && typeof o == 'object' && (o.constructor === Object || o.constructor === plainModule.constructor);
 };
 
 const comparison = (obj, needle) => obj === needle;
--- a/test/moduleuserapp/authentication.js
+++ b/test/moduleuserapp/authentication.js
@@ -1,0 +1,17 @@
+'use strict';
+
+Object.defineProperty(exports, '__esModule', {
+  value: true
+});
+
+exports.type = 'oauth2';
+
+exports.test = () => 'test';
+
+const oauth2Config = exports.oauth2Config = {};
+
+oauth2Config.autoRefresh = true;
+
+oauth2Config.authorizeUrl = {
+  url: 'https://example.com/auth/oauth/v2/authorize',
+};
--- a/test/moduleuserapp/index.js
+++ b/test/moduleuserapp/index.js
@@ -1,0 +1,31 @@
+'use strict';
+
+Object.defineProperty(exports, '__esModule', {
+  value: true
+});
+
+var _authentication = require('./authentication');
+
+function _interopRequireWildcard(obj) {
+  if (obj && obj.__esModule) {
+    return obj;
+  } else {
+    var newObj = {};
+    if (obj != null) {
+      for (var key in obj) {
+        if (Object.prototype.hasOwnProperty.call(obj, key)) {
+          newObj[key] = obj[key];
+        }
+      }
+    }
+    newObj.default = obj; return newObj;
+  }
+}
+
+var authentication = _interopRequireWildcard(_authentication);
+
+exports.default = {
+  authentication,
+};
+
+module.exports = exports['default'];  // eslint-disable-line
--- a/test/tools/resolve-method-path.js
+++ b/test/tools/resolve-method-path.js
@@ -5,11 +5,13 @@ require('should');
 const _ = require('lodash');
 
 const appDefinition = require('../userapp');
+const moduleAppDefinition = require('../moduleuserapp');
 const resolveMethodPath = require('../../src/tools/resolve-method-path');
 const schemaTools = require('../../src/tools/schema');
 
 describe('resolve-method-path', () => {
   const app = schemaTools.prepareApp(appDefinition);
+  const moduleApp = schemaTools.prepareApp(moduleAppDefinition);
 
   const oauthAppDef = _.extend({}, appDefinition);
   oauthAppDef.authentication = {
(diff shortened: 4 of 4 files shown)


ibolmo:
added @bryanhelmig since it was suggested, but also cc: @BrunoBernardino @bcooksey who might have better context if this might cause 🔥

bryanhelmig (commented):
Oh interesting -- I had no idea modules had different constructors... but it makes sense because _javascript_! 😛

src/tools/data.js
```diff
@@ -2,9 +2,10 @@
 
 const _ = require('lodash');
 const memoize = require('./memoize');
+const plainModule = require('./plain');
 
 const isPlainObj = (o) => {
```
BrunoBernardino, line 7: Can we do keep `isPlainObj` and add a `isPlainModule`, and then have the check for `isPlainObj || isPlainModule`? I think that's more clear than making an assumption and changing what `isPlainObj` really means.

BrunoBernardino (changes requested):
Awesome, didn't know about this limitation as well, thank you so much for this!

I've made a comment asking for a change in the code to increase readability.

Additionally, if possible and not too complicated, could you also add a test for it?

src/tools/data.js
ibolmo: Looks like `isPlainObj` is used in another `serializeApp` via the `recurseCleanFuncs`. From my understanding, we'll want to permit the `isPlainObj || isPlainModule` check there as well but really I think we _should_ augment `isPlainObj`. The `o.constructor` check is really unforgiving. 

Let me know, if you really think this should be separate and if the recurseCleanFuncs should have the `|| isPlainModule`

BrunoBernardino: You have the most context for it, so I'm definitely fine with keeping it like this.

BrunoBernardino (approved):

ibolmo:
While (painfully) trying to recreate this locally I discovered this was a problem on my end:

**Notice the `lib/index` (the transpiled version) was used (incorrectly)**

```js
/* eslint-env jest */
import { createAppTester, tools } from 'zapier-platform-core';
import replace from 'replace-in-file';

tools.env.inject();

import App from 'lib/index';

const appTester = createAppTester(App);


test('should refresh auth', async () => {
  const result = await appTester(App.authentication.oauth2Config.refreshAccessToken, {
    authData: {
      refresh_token: process.env.REFRESH_TOKEN,
    },
  });
```

**Now see the fixed version (uses source ES7) [see my note below, if you're curious about the import]**

```js
/* eslint-env jest */
import { createAppTester, tools } from 'zapier-platform-core';
import replace from 'replace-in-file';

tools.env.inject();

import App from 'app';

const appTester = createAppTester(App);


test('should refresh auth', async () => {
  const result = await appTester(App.authentication.oauth2Config.refreshAccessToken, {
    authData: {
      refresh_token: process.env.REFRESH_TOKEN,
    },
  });
```

Closing since this is working now without the fix. 😓 


---

If you're curious about how I could do `import App from 'app';` here's my babelrc:

```json
{
  "presets": [
    ["env", {"targets": {"node": 6}}]
  ],
  "plugins": [
    "add-module-exports",
    "transform-object-rest-spread",
    ["module-resolver", {
      "root": ["./src"],
      "alias": {
        "app": "./src",
        "lib": "./lib",
        "test": "./test",
      }
    }]
  ]
}
```

The [module-resolver](https://github.com/tleunen/babel-plugin-module-resolver) is awesome.

ibolmo:
So it turns out removing the `node_modules` and npm installing (thereby removing any local changes/fixes) brought back the error 😠 

I've spent quite a bit of time trying to reproduce it outside of the environment that this comes up (I'm using jest – which runs babel-node). 

At this point I'm thinking of merging without the additional test coverage. I'll keep trying to find a way to reproduce procedurally.

ibolmo:
So.. quite a bit of turn of events.

It turns out that modules as objects is an "exotic" occurrence ([babeljs slack](https://babeljs.slack.com/archives/C0DFJT81H/p1506715554000324) if curious). Yet, the babel folks agree that modules should behave like objects. I've left a request in jest to review why the constructor is different, but at this point that's moot. I created a separate repo: https://github.com/ibolmo/namespaced-objects-conundrum that reproduces the problem and babel as well as jest will take a look.

For now, the lodash `isPlainObject` is working as expected for both `babel-node` and `jest`. A discussion arose if lodash will in the future make an exception for modules in the isPlainObject check, but we don't have to worry about that until Node.JS has native modules. Here's the line that will fail when native modules kick in: https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L12059-L12060

Essentially they fall back to checking if the object's constructor is Object (like we did). But in native modules it'll be `[Function:  Module]`. 

Would be good to get your thoughts (@bryanhelmig  @BrunoBernardino) once more before proceeding.  

**My standing**: let's proceed using lodash' `isPlainObject` and if and when AWS gets native modules our test suite will throw a tantrum. Should we need to we can define our own checker like I did in the first commit. Likely, though, lodash (or the next gen lodash utility belt) will have implemented something.

bryanhelmig:
I think we should probably roll a custom `isPlainObject` that handles both? If AWS updates the node version overnight that'll be a nasty outage! 😱

ibolmo:
Alright we ended back to the original fix since I couldn't decouple the check for a new separate `isModuleObject`. I've planted the seed at lodash to consider working the problem, but going to default to action here. 

Should the spec decide to drop the `constructor` then it'll return `false` which is what we'd expect a plain object to be. Should the object be a module then the constructor should be the same. Jest has the curious behavior of  -- some how -- having a different constructor, but I've tested this separately (hard to do this in a mocha environment). 

Going to go ahead and merge.

PR #49, opened 2017-09-26
Use "application" discrete-type instead of "binary" as default content type.

charettes:
The "binary" subtype is not standard per RFC 2045 and causes issues with services expecting "application/octet-stream" instead.

See https://tools.ietf.org/html/rfc2045#section-5.1 for more details.


BrunoBernardino (approved):
Simple and straightforward, and with the complementary PR (that should ship first) at https://github.com/zapier/zapier/pull/13449 ! Awesome, thanks!

PR #50, opened 2017-10-04
Use unsafe data when logging to stdout

FokkeZB:
When running `zapier test --debug` it can be very [frustrating](https://secure.helpscout.net/conversation/442126165/677053/?folderId=20363) to have censored logs and IMHO as long as no hacker is peeking over your shoulder we should be fine logging unsafe data to the console.

I'm not sure if we can also use unsafe data for `options.logBuffer && data.log_type === 'console'` at https://github.com/zapier/zapier-platform-core/pull/50/files#diff-0936f4d83f660e8d3b1ba19e2225e4c6R114.

I'm also not sure how to best test this. Can we capture stdout in a test to verify it's unsafe?

### DIFF
--- a/src/tools/create-logger.js
+++ b/src/tools/create-logger.js
@@ -86,13 +86,13 @@ const sendLog = (options, event, message, data) => {
   data.log_type = data.log_type || 'console';
 
   const sensitiveBank = makeSensitiveBank(event);
-  message = truncate(cleaner.recurseReplaceBank(message, sensitiveBank));
-  data = cleaner.recurseReplaceBank(data, sensitiveBank);
-  data = dataTools.recurseReplace(data, truncate);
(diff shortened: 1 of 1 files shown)


bryanhelmig (approved):
Yeah makes perfect sense to me!

FokkeZB:
@BrunoBernardino any thoughts on a test for this before I merge?

BrunoBernardino:
Awesome stuff @FokkeZB ! I'd just add a test to make sure we're logging the data censored (if there isn't one already). The uncensored version is harder and less important to test for, since it should happen on console. Does that make sense?

FokkeZB:
@BrunoBernardino we have existing tests for that which still pass so :+1:
https://github.com/zapier/zapier-platform-core/blob/master/test/logger.js#L97

BrunoBernardino:
Awesome, thanks!

PR #51, opened 2017-10-19
Log http(s) requests for non-client library

BrunoBernardino:
This provides more logging even if `z.request` isn't used (like when SDKs are used), by patching the `http` library in node.

This was tested with Dropbox and the interpreter "full test" app (with different non-client http/request libraries).

[Relevant Trello Card](https://trello.com/c/uUo9ZCoU/193-cli-track-all-outgoing-requests-whatever-library-is-used).

### What this changes

- [x] Logs requests (safely) for any http(s) request happening, outside of `z.request`.
- [x] Uses the `user-agent: Zapier` http header to differentiate from client requests and non-client ones.
- [x] Adds the ability to test logging locally in other apps, via `process.env.LOGGING_TOKEN`.
- [x] Adds a `request_via_client: Boolean` data property for GL, to differentiate from the different logs.
- [x] Doesn't log the requests that are made to GL
- [x] Adds a couple of tests test for the `user-agent` header

### Log

---

**Log http(s) requests for non-client library** (f9111da)

Fixes https://trello.com/c/uUo9ZCoU/193-cli-track-all-outgoing-requests-whatever-library-is-used

This also adds a "user-agent: Zapier" header, which is used to differentiate from client requests and non-client ones.

It also adds the ability to test logging locally in other apps, via the `process.env.LOGGING_TOKEN`

---

**Removing memory leak and adding more tests** (c84f6d9)

---

**Don't set encoding.** (f846d6e)

### DIFF
--- a/src/http-middlewares/before/prepare-request.js
+++ b/src/http-middlewares/before/prepare-request.js
@@ -17,6 +17,7 @@ const sugarBody = (req) => {
   // move into the body as raw, set headers for coerce, merge to work
 
   req.headers = req.headers || {};
+  req.headers['user-agent'] = 'Zapier';
 
   if (!req.body && req.form) {
     req.body = req.form;
--- a/src/http-middlewares/before/prepare-request.js
+++ b/src/http-middlewares/before/prepare-request.js
@@ -17,7 +17,6 @@ const sugarBody = (req) => {
   // move into the body as raw, set headers for coerce, merge to work
 
   req.headers = req.headers || {};
-  req.headers['user-agent'] = 'Zapier';
 
   if (!req.body && req.form) {
     req.body = req.form;
--- a/src/tools/create-http-patch.js
+++ b/src/tools/create-http-patch.js
@@ -55,8 +55,7 @@ const createHttpPatch = (event) => {
           logger(`${logData.response_status_code} ${logData.request_method} ${logData.request_url}`, logData);
         };
 
-        response.setEncoding('utf8');
-        response.on('data', (chunk) => responseBody.push(chunk));
+        response.on('data', (chunk) => responseBody.push(chunk.toString()));
         response.on('end', logResponse);
         response.on('error', logResponse);
 
--- a/src/tools/create-http-patch.js
+++ b/src/tools/create-http-patch.js
@@ -1,0 +1,69 @@
+const _ = require('lodash');
+const constants = require('../constants');
+
+const createHttpPatch = (event) => {
+  const createLogger = require('./create-logger');
+  const logBuffer = [];
+  const logger = createLogger(event, {logBuffer});
+
+  const httpPatch = (object) => {
+    const originalRequest = object.request;
+
+    // Proxy the request method
+    object.request = (options, callback) => {
+      const requestUrl = options.url || options.href || (`${options.protocol || 'https://'}${options.host}${options.path}`);
+      const logger_url = process.env.LOGGING_ENDPOINT || constants.DEFAULT_LOGGING_HTTP_ENDPOINT;
+
+      // Ignore logger requests
+      if (requestUrl.indexOf(logger_url) !== -1) {
+        return originalRequest(options, callback);
+      }
+
+      // Ignore requests made via the request client
+      if (_.get(options.headers, 'user-agent', []).indexOf('Zapier') !== -1) {
+        return originalRequest(options, callback);
+      }
+
+      // Proxy the callback to get the response
+      const newCallback = function(response) {
+        const requestHeaders = _.map(Object.keys(options.headers || {}), (header) => `${header}: ${options.headers[header]}`);
+        const responseHeaders = _.map(Object.keys(response.headers || {}), (header) => `${header}: ${response.headers[header]}`);
+        const responseBody = [];
+
+        const logResponse = () => {
+          // Prepare data for GL
+          const logData = {
+            log_type: 'http',
+            request_type: 'devplatform-outbound',
+            request_url: requestUrl,
+            request_method: options.method || 'GET',
+            request_headers: requestHeaders.join('\n'),
+            request_data: options.body || '',
+            request_via_client: false,
+            response_status_code: response.statusCode,
+            response_headers: responseHeaders.join('\n'),
+            response_content: responseBody.join(''),
+          };
+
+          logger(`${logData.response_status_code} ${logData.request_method} ${logData.request_url}`, logData);
+        };
+
+        response.setEncoding('utf8');
+        response.on('data', (chunk) => responseBody.push(chunk));
+        response.on('end', logResponse);
+        response.on('error', logResponse);
+
+        // If there was a callback, call it now
+        if (_.isFunction(callback)) {
+          callback.apply(this, arguments);
+        }
+      };
+
+      return originalRequest(options, newCallback);
+    };
+  };
+
+  return httpPatch;
+};
+
+module.exports = createHttpPatch;
--- a/src/tools/create-http-patch.js
+++ b/src/tools/create-http-patch.js
@@ -9,6 +9,13 @@ const createHttpPatch = (event) => {
   const httpPatch = (object) => {
     const originalRequest = object.request;
 
+    // Avoids multiple patching and memory leaks (mostly when running tests locally)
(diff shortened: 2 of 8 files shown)


src/http-middlewares/before/prepare-request.js
```diff
@@ -17,6 +17,7 @@ const sugarBody = (req) => {
   // move into the body as raw, set headers for coerce, merge to work
 
   req.headers = req.headers || {};
+  req.headers['user-agent'] = 'Zapier';
```
FokkeZB: The user **is** able to override this right?

src/tools/create-lambda-handler.js
```diff
@@ -25,6 +26,11 @@ const loadApp = (appRawOrPath) => {
 const createLambdaHandler = (appRawOrPath) => {
 
   const handler = (event, context, callback) => {
+    // Adds logging for _all_ kinds of http(s) requests, no matter the library
+    const httpPatch = createHttpPatch(event);
+    httpPatch(require('http'));
+    // httpPatch(require('https')); // NOTE: This was double-logging
```
FokkeZB: Yup, uses `http` under the hood.

FokkeZB (approved):

src/tools/create-http-patch.js
```diff
@@ -0,0 +1,69 @@
...
+
+        const logResponse = () => {
+          // Prepare data for GL
+          const logData = {
+            log_type: 'http',
+            request_type: 'devplatform-outbound',
+            request_url: requestUrl,
+            request_method: options.method || 'GET',
+            request_headers: requestHeaders.join('\n'),
+            request_data: options.body || '',
```
bcooksey, line 48: Might make sense to check if this was a file and do something like <probably binary data> instead of the full body.

src/tools/create-lambda-handler.js
bcooksey: Guess just nuke it then!

bcooksey (commented):
Interesting take on this @BrunoBernardino, I like it!

As part of this effort, it might make sense to add some documentation on how devs can get logging working on their own. Basically show off what tooling we provide and how they might hook into it. My thought is that this is a great start, and could end up covering lots of use cases. I could see some though where it's not an http protocol, or doesn't use the node standard lib, or maybe a dev wants to wire in great logging to an SDK and overwrite our default one. I think it would only take a code snippet or two and a bit of explanation to show what logging call to make to do that.

src/http-middlewares/before/prepare-request.js
BrunoBernardino: Yes, but I'll add that to the headers merging test just to make sure.

src/tools/create-http-patch.js
BrunoBernardino: I didn't want to do it here because the `createLogger` takes care of truncating and similar "cleaning" (like obfuscating sensitive data), so it'd effectively be almost duplicating work in a very quick sequential way.

If you think it's still a nice improvement, I'm happy to add it here.

BrunoBernardino:
Thanks @bcooksey ! Can you please elaborate on that logging thought?

As a dev, I can require a logging lib and call it before/after requests, or at specific parts of the app. We don't expose our logger (especially because it's specific to _our_ needs and logging tools), and if we did, I can see how that'd be more confusing (to create a logger instance with some `ENV` data, then log in a specific format), rather than just using their own logging lib.

Or are you asking for me to add such snippets to the CLI README (with an external logging tool like papertrail, not using/overriding our own logging)?

FokkeZB:
@BrunoBernardino why wouldn't we log these requests to Graylog?

BrunoBernardino:
@FokkeZB to avoid circular/infinite loops 😄

bcooksey:
@BrunoBernardino I was actually thinking about exposing our logger. So essentially, there would be three options:

1) Use `z.requests` - logged automatically
2) Use a library that uses Node's http lib - logged automatically
3) Docs to show devs how they can do something like:
   ```
   const logger = require('zapier-logger'); // I didn't look to see what the actual package is
   logger.log('Sending note to Evernote via Thrift', {noteSize: 1322, notebookId: 3})
   ```

Would that third option make sense?

BrunoBernardino:
I see... How is that different from `z.console.log()`? I don't think we should expose that "without any bundle/z info" because then it can't do the obfuscation, though it wouldn't happen with an external logging library as well.

bcooksey:
@BrunoBernardino Ah right. I was thinking we blocked those logs for public apps, but they get through, so that would be enough!

src/tools/create-http-patch.js
bcooksey: Nope, if it is covered elsewhere that is fine

src/tools/create-http-patch.js
```diff
@@ -0,0 +1,76 @@
...
+            response_content: responseBody.join(''),
+          };
+
+          logger(`${logData.response_status_code} ${logData.request_method} ${logData.request_url}`, logData);
+        };
+
+        response.setEncoding('utf8');
+        response.on('data', (chunk) => responseBody.push(chunk));
+        response.on('end', logResponse);
+        response.on('error', logResponse);
```
bcooksey: I don't think we can safely attach all of these. I wrote [a gist](https://gist.github.com/bcooksey/2dc38ffd18f0eda87fb1f79d4c56fed2) to demonstrate some behaviors and found this:

1) `setEncoding()` swaps the incoming values to the data listener from buffers to strings, which is probably not a nice thing to do to devs
2) `setEncoding()` will, obviously, encode the data, so if binary data being transferred, devs would have to decode it to get back to what they want
3) Adding a data handler [swaps the stream mode](https://nodejs.org/dist/latest-v6.x/docs/api/stream.html#stream_two_modes) from paused to flowing. This can cause `.read()` to return no data. As far as how much we should care, seems like an app will end up doing a `.pipe()` or something and get it into flow mode anyway, but it is a nasty edge to force on folks. If we knew consuming data via `.read()` was rare, I'd be fine ignoring this point.

bcooksey (changes requested):
Some things to consider with how we tap into the response.

src/tools/create-http-patch.js
BrunoBernardino: Wow, nice catches! I'll see if I can tap into that without changing data (will use your gist).

If I can't, I'll remove that and we'll log without the response content (still better than nothing). If it becomes something we _really_ want, we can then try to look into ways of tapping into it more safely, maybe.

FokkeZB: You may also want to look into how NewRelic wraps `http.request` and friends. Pretty sure they've made sure not to break it: https://github.com/newrelic/node-newrelic/blob/master/lib/instrumentation/core/http.js#L579-L606

BrunoBernardino: Thanks @FokkeZB. After looking through it and trying it out for a while, it seems their wrapper libraries/methods are very complex and generic (and the request/response parts use the agent and eventemitters, which, while _probably better_, seem like a major overkill for what we're going for here).

I was able to confirm no app in CLI is currently calling anything `.read()`, so I think we're safe on @bcooksey 's third concern. I ditched the `setEncoding` and that seems to still be working fine (worst case scenario we might have some encoding issues on GL, and we can try to tweak that if so, later).

Let me know if that works fine.

src/tools/create-http-patch.js
```diff
@@ -0,0 +1,75 @@
...
+            request_via_client: false,
+            response_status_code: response.statusCode,
+            response_headers: responseHeaders.join('\n'),
+            response_content: responseBody.join(''),
+          };
+
+          logger(`${logData.response_status_code} ${logData.request_method} ${logData.request_url}`, logData);
+        };
+
+        response.on('data', (chunk) => responseBody.push(chunk.toString()));
```
bcooksey, line 58: Cool, this way we'll still get utf8 strings being logged, but it won't mess with any other listeners on the `'data'` event.

bcooksey (approved):
Alright, I think this is looking good now. I'm still a bit unsure on the use of `read()`. It's not in app code, but what about sdk's? However, Node docs push folks to use `.pipe`. If we do run into an issue, we may be able to have an app declare it wants to disable the http patch some way.

PR #52, opened 2017-11-14
Encode form requests with + instead of %20 by default

BrunoBernardino:
Most apps and the specs (read https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent on application/x-www-form-urlencoded) expect a plus sign instead of %20 in this situation.

Tests were also added. This will probably require a breaking release.

Fixes https://github.com/zapier/zapier-platform-cli/issues/161

Thoughts @bcooksey ?

### DIFF
--- a/src/http-middlewares/before/prepare-request.js
+++ b/src/http-middlewares/before/prepare-request.js
@@ -39,8 +39,7 @@ const coerceBody = (req) => {
 
   // auto coerce form if header says so
   if (contentType === FORM_TYPE && req.body && !_.isString(req.body)) {
-    // TODO: use const FormData = require('form-data'); instead?
-    req.body = querystring.stringify(req.body);
+    req.body = querystring.stringify(req.body).replace(/%20/g, '+');
   }
 
   if (isStream(req.body)) {
--- a/test/create-request-client.js
+++ b/test/create-request-client.js
@@ -39,8 +39,6 @@ describe('request client', () => {
       .then(responseBefore => {
         const response = JSON.parse(JSON.stringify(responseBefore));
 
-        console.log(response.request.headers);
-
         should(response.request.headers['user-agent']).eql(undefined);
         response.request.headers['User-Agent'].should.eql('Zapier!');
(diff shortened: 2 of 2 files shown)


eliangcs (approved):
I didn't pull this because this looks straightforward enough. My only suggestion is that we should add a test making sure `+` is not encoded to `%2B`, which can happen if you do `.replace(' ', '+')` first then `querystring.stringify()`.

BrunoBernardino:
@eliangcs that's a good point. I'm not sure we wouldn't want those converted, though. We're just trying to keep the spaces converted to `+`, the spec doesn't say anything about not converting `+`, especially because if we don't do that conversion, there's no way of sending an actual `+`, then.

Does that make sense, or am I missing something?

eliangcs:
@BrunoBernardino I just realized that I was confused. You're right: `+` should be encoded. Otherwise, there's no way to send an actual `+`. Please ignore my last comment.

src/http-middlewares/before/prepare-request.js
```diff
@@ -39,8 +39,7 @@ const coerceBody = (req) => {
 
   // auto coerce form if header says so
   if (contentType === FORM_TYPE && req.body && !_.isString(req.body)) {
-    // TODO: use const FormData = require('form-data'); instead?
-    req.body = querystring.stringify(req.body);
+    req.body = querystring.stringify(req.body).replace(/%20/g, '+');
```
bcooksey, line 42: Sanity check that node doesn't have a built-in lib to do the encoding the way we want, vs having to do this find and replace?

BrunoBernardino: Nice idea! Unfortunately there isn't, because we want to do it for everything _but_ spaces. The [`options` for that method](https://nodejs.org/api/all.html#querystring_querystring_stringify_obj_sep_eq_options) only include sending a function that's `encodeURIComponent` by default.

PR #54, opened 2017-11-17
Replace _.truncate, save memory.

### LINKED ISSUE #53: Some of Lodash's string functions may be consuming a lot of memory on AWS Lambda
## Description

While working on file uploads with a CLI App, we've seen regular errors from Lambda and our function invocations dying. Recently, I enabled logging on one of these functions and found this very brief JS stacktrace:

That `unicodeToArray` got me curious. Looking at [Lodash source](https://github.com/lodash/lodash/blob/d10b44bdef2b735865c493fbaec8687ec8998aef/.internal/unicodeToArray.js#L38), we see this is indeed running a regular expression on whatever comes into it. Turns out this function is used by `stringToArray`, which is called by lots of Lodash functions in the trim, pad, and truncate realm.

With this bit of a hint, I went hunting and was able to find [this use of truncate](https://github.com/zapier/zapier-platform-core/blob/5fed875579a82f7edc8a56cc7ff014b5503fc098/src/tools/create-logger.js#L12) by our logger.

My theory is this: When a large file like a 25 MB movie gets passed through lambda, our logger is running `_.truncate()` on the raw binary string of data. Underhood, Lodash is running that regular expression on the binary string and it's causing massive amounts of memory usage.

## Replicating

This is the hard part. I had a consistent failure scenario, made a tweak locally to our logger, uploaded it in a test app. Errors went away! Then I reverted the fix and continued to see success, so I'm not positive what's happening anymore.

BrunoBernardino:
Fixes #53

Memory usage for truncating a 25MB file went from a 900-1000MB spike to a **130-160MB** spike! I also confirmed the output matched in both situations (`_.truncate()` actually just "appended" the `omission`, while now we make sure the string doesn't go over the defined length _including_ the `suffix`, but I think it like it better this way).

Also, plenty of tests were added for it, and a few other places that were using `_.truncate()` were updated too.

### DIFF
--- a/src/tools/data.js
+++ b/src/tools/data.js
@@ -128,6 +128,20 @@ const flattenPaths = (data, sep) => {
   return out;
 };
 
+// A simpler, and memory-friendlier version of _.truncate()
+const simpleTruncate = (string, length, suffix) => {
+  if (!string || !string.toString || string.toString().length === 0) {
+    return '';
+  }
+
+  if (string.toString().length > length) {
+    const cutoff = (suffix ? length - suffix.length : length);
+    return string.toString().substr(0, cutoff) + (suffix ? suffix : '');
+  }
+
+  return string;
+};
+
 module.exports = {
   isPlainObj,
   findMapDeep,
@@ -136,5 +150,6 @@ module.exports = {
   jsonCopy,
   deepFreeze,
   recurseReplace,
-  flattenPaths
+  flattenPaths,
+  simpleTruncate,
 };
--- a/src/tools/data.js
+++ b/src/tools/data.js
@@ -130,13 +130,17 @@ const flattenPaths = (data, sep) => {
 
 // A simpler, and memory-friendlier version of _.truncate()
 const simpleTruncate = (string, length, suffix) => {
-  if (!string || !string.toString || string.toString().length === 0) {
+  if (!string || !string.toString) {
     return '';
   }
 
-  if (string.toString().length > length) {
+  const finalString = string.toString();
+
+  if (finalString.length === 0) {
+    return '';
+  } else if (finalString.length > length) {
     const cutoff = (suffix ? length - suffix.length : length);
-    return string.toString().substr(0, cutoff) + (suffix ? suffix : '');
+    return finalString.substr(0, cutoff) + (suffix ? suffix : '');
   }
 
   return string;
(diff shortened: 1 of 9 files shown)


src/tools/data.js
```diff
@@ -128,6 +128,20 @@ const flattenPaths = (data, sep) => {
...
 
+// A simpler, and memory-friendlier version of _.truncate()
+const simpleTruncate = (string, length, suffix) => {
+  if (!string || !string.toString || string.toString().length === 0) {
+    return '';
+  }
+
+  if (string.toString().length > length) {
+    const cutoff = (suffix ? length - suffix.length : length);
+    return string.toString().substr(0, cutoff) + (suffix ? suffix : '');
```
bcooksey: Instead of invoking `toString()` three times, could we put that in a variable?

test/misc-tools.js
```diff
@@ -55,6 +55,30 @@ describe('Tools', () => {
...
+      {value: () => {}, length: 5, suffix: '...', expected: '()...'},
+      {value: {yeah: true}, length: 8, suffix: '...', expected: '[obje...'},
+      {value: 'Something', length: 5, suffix: undefined, expected: 'Somet'},
+      {value: 'Something', length: 5, suffix: '...', expected: 'So...'},
+      {value: new Buffer('Something'), length: 7, suffix: ' [...]', expected: 'S [...]'},
+      {value: 'Something', length: 0, suffix: '...', expected: '...'},
+      {value: 'Something', length: 8, suffix: '...', expected: 'Somet...'},
+      {value: 'Something', length: 9, suffix: '...', expected: 'Something'},
+      {value: 'Something', length: 15, suffix: '...', expected: 'Something'},
+    ];
```
bcooksey: Phenomenal test!

bcooksey (approved):
One suggestion for efficiency, but otherwise this looks good.

Also, when digging into Lodash I recall them jumping over hoops to do special handling with unicode. I have not been able to figure why though, and can't find a test suite in the source for clues

ibolmo:
(looking at this via phone) The Unicode jumping, I believe, is that the
length of the string varies. JS expects utf-8 but utf-16 strings will be
longer length.

-- 
Olmo Maldonado
Partners and Platform Engineer – Zapier.com

bcooksey:
Also cc @kola-er. @BrunoBernardino was able to get a patch in, so you should be able to use this once it merges for DropBox!

kola-er:
@bcooksey Yeah, I've been doing something similar locally. I just pulled and tested with it; the weirdness in memory usage is gone. Thanks @BrunoBernardino

test/misc-tools.js
BrunoBernardino: I'll add some unicode tests.

PR #55, opened 2017-11-22
Don't add content-type if there's one

BrunoBernardino:
Fixes a problem where defining a content-type header with a different case (like 'Content-Type') and a non-string `request.body` would cause another `content-type` (lowercase) header to be set.

Also adds tests.

[Trello Card](https://trello.com/c/u3vUzz1N/209-cli-the-header-content-type-does-not-replace-the-default-header-content-type).


kola-er (approved):
Extensive test cases. Good to go.

PR #56, opened 2017-12-12
Ignore 'tags' when building

eliangcs:
'tags' file is generated by ctags and used to help navigating code to ease development. It was accidentally packed into 4.1.0 package by a person (me) who uses ctags. It should be removed from the package.

Related: zapier/zapier-platform-cli#165

P.S. For the longer term, I think it's best to use Travis CI to help us release production builds to avoid this kind of problem.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -29,4 +29,5 @@ find . -print | \
     grep -v "\.md" | \
     grep -v "\.sh" | \
     grep -v "\.zip" | \
+    grep -v "^./tags$" | \
     zip $1 -@
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -29,5 +29,5 @@ find . -print | \
     grep -v "\.md" | \
     grep -v "\.sh" | \
     grep -v "\.zip" | \
-    grep -v "^./tags$" | \
+    grep -v "^\./tags$" | \
     zip $1 -@


xavdid:
I'm missing a little context here - could you talk about what this builds? Does the core get shipped separately to lambda? 

I'm a little concerned with how many exclusions we're adding. What happens if we add a hypothetical `hashtags.js` that's part of the lib. It would get skipped, right?

Maybe we should build inclusively instead? Explicitly add `src`, etc.

eliangcs:
@xavdid Try `npm install zapier-platform-core@4.1.0` and `cd node_modules/zapier-platform-core`. You'll see a 22MB file named 'tags'. It was not there until 4.1.0.

And I'm using regex pattern `^./tags$`, so it should exclude `./tags` and `./tags` only. Other things like `./tags/123`, `./123/tags` and `./hashtags.js` won't get skipped.

xavdid:
That's fair. This fixes the immediate problem, i'm just not sure how it helps the longterm issue of 

1. accidentally add something
2. realize
3. update build script
4. push smaller version. 

So yeah on the surface this is fine, it would just be cool to get in front of that ^

bcooksey (approved):
Whoops, def don't want to include that in the builds.

To @xavdid's point about avoiding this in the future, I think @eliangcs idea of travis would be a good one. We already have an account. Would it be tough to setup?

xavdid:
very easy! https://docs.travis-ci.com/user/deployment/npm/

PR #58, opened 2017-12-15
Prettier

xavdid:
same as schema, the good stuff is in the first commit


xavdid:
eslint and prettier are fighting about spacing. prettier should take precedent, so i'll see about making those play nicely. there's an eslint config for that.

eliangcs (approved):
Pulled and tested. Works great!

PR #59, opened 2017-12-19
Fix TypeError on requestUrl.indexOf

eliangcs:
Encountered the following error when I was working on zapier/zapier-platform-cli#208. 

```
  1) Searches - Full Search
       should get an object:
     TypeError: requestUrl.indexOf is not a function
      at Object.object.request (node_modules/zapier-platform-core/src/tools/create-http-patch.js:25:22)
      at Request.start (node_modules/request/request.js:748:32)
      at Request.end (node_modules/request/request.js:1512:10)
      at end (node_modules/request/request.js:561:14)
      at Immediate.<anonymous> (node_modules/request/request.js:575:7)
      at Function.module.exports.loopWhile (node_modules/deasync/index.js:72:22)
      at node_modules/deasync/index.js:44:19
      at Object.request (node_modules/zapier-platform-legacy-scripting-runner/z.js:97:22)
      at Object.full_search_search (scripting.js:97:26)
      at e (node_modules/zapier-platform-legacy-scripting-runner/index.js:108:33)
      at Object.runEvent (node_modules/zapier-platform-legacy-scripting-runner/index.js:32:35)
      at getList (searches/full_search.js:23:6)
      at execute (node_modules/zapier-platform-core/src/execute.js:68:12)
      at node_modules/zapier-platform-core/src/create-command-handler.js:26:12
      at Object.beforeMiddleware.then.newInput (node_modules/zapier-platform-core/src/middleware.js:91:24)
      at bound (domain.js:280:14)
      at Object.runBound (domain.js:293:12)
```

Turned out that `options.url` is an object as opposed to a string in this failed test case. According to [http.request's doc](https://nodejs.org/docs/latest-v6.x/api/http.html#http_http_request_options_callback), `options` is either a string (URL) or an object (request options). It doesn't say anything about `options.url`. I'm not sure why the previous implementation did `options.url || options.href`, but here we go.

### DIFF
--- a/src/tools/create-http-patch.js
+++ b/src/tools/create-http-patch.js
@@ -18,10 +18,20 @@ const createHttpPatch = event => {
 
     // Proxy the request method
     object.request = (options, callback) => {
-      const requestUrl =
-        options.url ||
-        options.href ||
-        `${options.protocol || 'https://'}${options.host}${options.path}`;
+      // `options` can be an object or a string. If options is a string, it is
+      // automatically parsed with url.parse().
+      // See https://nodejs.org/docs/latest-v6.x/api/http.html#http_http_request_options_callback
+      let requestUrl;
+      if (typeof options === 'string') {
+        requestUrl = options;
+      } else {
+        requestUrl =
+          options.href ||
+          `${(options.protocol || 'https:') + '//'}${options.host}${
+            options.path
+          }`;
+      }
+
       const logger_url =
         process.env.LOGGING_ENDPOINT || constants.DEFAULT_LOGGING_HTTP_ENDPOINT;
 
--- a/src/tools/create-http-patch.js
+++ b/src/tools/create-http-patch.js
@@ -27,9 +27,7 @@ const createHttpPatch = event => {
       } else {
         requestUrl =
           options.href ||
-          `${(options.protocol || 'https:') + '//'}${options.host}${
-            options.path
-          }`;
+          `${options.protocol || 'https:'}//${options.host}${options.path}`;
       }
 
       const logger_url =


grajagandev:
I love Zapier - great product. Would you be interested in software that automatically finds TypeErrors like this? I am building [Fuzz Stati0n](https://fuzzstati0n.com/newsletter-dec-17.html) to do that (free for OSS) - please take a look and consider signing up for our newsletter to keep in touch.

src/tools/create-http-patch.js
```diff
@@ -18,10 +18,20 @@ const createHttpPatch = event => {
 
     // Proxy the request method
     object.request = (options, callback) => {
-      const requestUrl =
-        options.url ||
```
xavdid, line 22: I believe this isn't re-added anywhere, right? So this will break existing code that is passing in objects with a `.url` property, which may not be intended.

src/tools/create-http-patch.js
```diff
@@ -18,10 +18,20 @@ const createHttpPatch = event => {
...
+      // `options` can be an object or a string. If options is a string, it is
+      // automatically parsed with url.parse().
+      // See https://nodejs.org/docs/latest-v6.x/api/http.html#http_http_request_options_callback
+      let requestUrl;
+      if (typeof options === 'string') {
+        requestUrl = options;
+      } else {
+        requestUrl =
+          options.href ||
+          `${(options.protocol || 'https:') + '//'}${options.host}${
```
xavdid: super minor, but you can have the `//` in the template string itself instead of doing the addition operation:

`${(options.protocol || 'https:')}//${options.host}...`

xavdid (commented):
If the `.url` thing is a breaking change, we should fix it. if not, you're probably good to go. the string thing is super not worth its own commit.

src/tools/create-http-patch.js
eliangcs: If I understand the code correctly, `object.request` is to patch Node's [`http.request`](https://nodejs.org/docs/latest-v6.x/api/http.html#http_http_request_options_callback) method, which does not accept an object with a `.url` property:

```
> let http = require('http');
undefined
> http.request({url:'http://httpbin.org/get'});
...
> Error: connect ECONNREFUSED 127.0.0.1:80
    at Object.exports._errnoException (util.js:1018:11)
    at exports._exceptionWithHostPort (util.js:1041:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1090:14)
```

 If the point is to follow http.request's spec 100%, then we should not accept an object with a `.url` property here neither. But I'm not sure if any existing code or library is doing that (passing an object with a `.url` property). Any clues about that?

xavdid: yeah, fair deal! should be safe to go then.

src/tools/create-http-patch.js
eliangcs: Good suggestion. Fixed in e754267.

PR #60, opened 2017-12-21
don't use buffer constructor

xavdid:
Cousin PR of [this one](https://github.com/zapier/zapier-platform-cli/pull/213).

There's also a change to use `Buffer.isBuffer` instead of `_.isBuffer`, which uses the former under the hood. Consistency seems like a good thing.

### DIFF
--- a/integration-test/integration-test.js
+++ b/integration-test/integration-test.js
@@ -24,7 +24,7 @@ const runLambda = event => {
         return reject(err);
       }
 
-      const logs = new Buffer(data.LogResult, 'base64').toString();
+      const logs = new Buffer.from(data.LogResult, 'base64').toString();
       console.log('\n=== LOGS ===\n', logs, '\n===\n');
 
       const response = data.Payload ? JSON.parse(data.Payload) : data;
--- a/integration-test/integration-test.js
+++ b/integration-test/integration-test.js
@@ -24,7 +24,7 @@ const runLambda = event => {
         return reject(err);
       }
 
-      const logs = new Buffer.from(data.LogResult, 'base64').toString();
+      const logs = Buffer.from(data.LogResult, 'base64').toString();
       console.log('\n=== LOGS ===\n', logs, '\n===\n');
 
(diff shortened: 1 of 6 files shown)


integration-test/integration-test.js
```diff
@@ -24,7 +24,7 @@ const runLambda = event => {
         return reject(err);
       }
 
-      const logs = new Buffer(data.LogResult, 'base64').toString();
+      const logs = new Buffer.from(data.LogResult, 'base64').toString();
```
eliangcs: `new` should be removed here.

test/create-request-client.js
```diff
@@ -121,7 +121,7 @@ describe('request client', () => {
     request({
       method: 'POST',
       url: 'http://zapier-mockbin.herokuapp.com/request', // httpbin doesn't handle chunked anything :-(
-      body: new Buffer('hello world this is a cat (=^..^=)')
+      body: new Buffer.from('hello world this is a cat (=^..^=)')
```
eliangcs: `new` should be removed.

test/tools/file-stasher.js
```diff
@@ -46,7 +46,7 @@ describe('file upload', () => {
     mocky.mockRpcCall(mocky.fakeSignedPostData);
     mocky.mockUpload();
 
-    const file = new Buffer('hello world this is a buffer of text');
+    const file = new Buffer.from('hello world this is a buffer of text');
```
eliangcs: Same here.

eliangcs (changes requested):
Instead of `new Buffer.from()`, I think you meant to write `Buffer.from()`. But strangely enough, `new Buffer.from()` actually works like `Buffer.from()`:

```
> new Buffer.from('aa')
<Buffer 61 61>
> Buffer.from('aa')
<Buffer 61 61>
```

xavdid:
ah, you're super right! i had a little bit of tunnel vision with regards to changing the constructor to `.from`, haha. I've updated!

eliangcs (approved):

PR #61, opened 2018-01-04
Fix http.request patch so it accepts options.url

eliangcs:
#59 broke the existing code. Some HTTP requests don't work because of that. This PR fixes the problem by looking at `options.url` in `http.request`'s patch, bringing back the previous behavior.


eliangcs:
@xavdid [you were right](https://github.com/zapier/zapier-platform-core/pull/59#discussion_r157849687). It did break the existing code. I'm merging this without code review as this is 🔥.

PR #62, opened 2018-01-16
For gzip encoded response, unzip before logging response.

kishorenc:
This fixes https://github.com/zapier/zapier/issues/14888

The issue is on all gzip encoded responses. Here's a recent Graylog entry with bad response content: https://logs.internal.zapier.com/messages/graylog_1833/c0394913-faaf-11e7-be8a-0e0ca1cbbd4e

I would like some help in testing this locally with an actual CLI app to verify the fix on Graylog (probably on staging).

### DIFF
--- a/src/tools/create-http-patch.js
+++ b/src/tools/create-http-patch.js
@@ -1,3 +1,4 @@
+const zlib = require('zlib');
 const _ = require('lodash');
 const constants = require('../constants');
 
@@ -57,9 +58,10 @@ const createHttpPatch = event => {
           Object.keys(response.headers || {}),
           header => `${header}: ${response.headers[header]}`
         );
-        const responseBody = [];
 
-        const logResponse = () => {
+        const chunks = [];
+
+        const sendToLogger = (responseBody) => {
           // Prepare data for GL
           const logData = {
             log_type: 'http',
@@ -71,7 +73,7 @@ const createHttpPatch = event => {
             request_via_client: false,
             response_status_code: response.statusCode,
             response_headers: responseHeaders.join('\n'),
-            response_content: responseBody.join('')
+            response_content: responseBody
           };
 
           logger(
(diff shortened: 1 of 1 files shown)


xavdid (approved):
Cool! Thanks for doing this. 

As far as testing goes, I whipped up a sample app what gets gzipped data from google (and doesn't require auth). It's pushed as `App2174CLIAPI` and the source is here: https://gist.github.com/xavdid/c2e87a363bbd92a2c11eea147d558323 (our cli code syncing is busted right now). I've confirmed the output from this call is garbled when tests are run locally, so we should be able to test the fix in GL. 

Side note, local always just spits out the response, but it could be worthwhile doing the same fix in CLI if it's useful to see the output when testing with `--debug`. It seems like it hasn't come up as a feature request yet though, so I wouldn't worry about it for now.

Anyway, it'll be tough to test this in staging. During build, we check the version against known platform version ([source](https://github.com/zapier/zapier/blob/746d5df2f69ff153da65a59a774c6d9df9cb4eee/developer_cli/api_resources.py#L1380-L1384)) and though npm can uses branches, I can't imagine we allow that. Maybe talk @eliangcs into releasing this as a [beta](https://docs.npmjs.com/getting-started/using-tags), which might trick whatever we use to verify server side that this is allowed. 

If that doesn't work, we could push a version of zapier to staging that forgoes the version verification and deploy that way, probably.

kishorenc:
@xavdid Thanks for the review. What I meant by testing on staging is, primarily pointing the `LOGGING_ENDPOINT` to Graylog staging and just verifying that response content in the logs don't show up as garbled anymore. If you have already verified that, then I think we can merge.

xavdid:
Ah, that does sound way easier! 

I've got the stuff to test. I guess all I'd need from you is the endpoint to use and potentially a `token`? `create-logger.js`[this line](https://github.com/zapier/zapier-platform-core/blob/master/src/tools/create-logger.js#L118) which means nothing from local gets logged.

PR #65, opened 2018-02-21
Support test frameworks other than mocha

xavdid:
Mocha is _fine_, but there's a lot of other ways to test. I added another common one, plus checking for an env variable. 

Any test framework can be used by prepending the variable to the command in `package.json`:

```
"test": "NODE_ENV=testing my_test_framework"
```


eliangcs (approved):
I didn't pull this. Looks straightforward!

PR #67, opened 2018-02-22
PDE-83: Check for object in create

mojowen:
## Type of Changes

|   | Type |
| ------------- | ------------- |
|    | :bug: Bug fix  |
|   | :sparkles: New feature |
| ✓  | :hammer: Refactoring  |
|   | :link: Update dependencies |
|   | :scroll: Docs |

## Related Issue
[PDE-83](https://zapierorg.atlassian.net/browse/PDE-83)

## Before / Current Behavior
Currently only validates that create returns a non-array

## After / New Behavior
Validates that the return value is an object

## TODOs

* [x] Add tests
* [x] All tests passing

### DIFF
--- a/src/checks/trigger-has-unique-ids.js
+++ b/src/checks/trigger-has-unique-ids.js
@@ -23,7 +23,7 @@ const triggerHasUniqueIds = {
     });
 
     if (doubleId !== undefined) {
-      return [`Got a two or more results with the id of ${doubleId}`];
+      return [`Got a two or more results with the id of "${doubleId}"`];
     }
     return [];
   }
--- a/src/checks/trigger-is-object.js
+++ b/src/checks/trigger-is-object.js
@@ -22,7 +22,9 @@ const triggerIsObject = {
 
     if (nonObjectResult !== undefined) {
       const repr = simpleTruncate(JSON.stringify(nonObjectResult), 50);
-      return [`Got a result missing that was not an object (${repr})`];
+      return [
+        `Got a non-object result in the array, expected only objects (${repr})`
+      ];
     }
     return [];
   }
--- a/src/checks/create-is-object.js
+++ b/src/checks/create-is-object.js
@@ -8,14 +8,14 @@ const isCreate = require('./is-create');
 /*
   Makes sure the results are all objects.
 */
-const createIsSingle = {
-  name: 'createIsSingle',
+const createIsObject = {
+  name: 'createIsObject',
   shouldRun: isCreate,
   run: (method, results) => {
-    if (_.isArray(results)) {
+    if (!_.isPlainObject(results)) {
       const repr = simpleTruncate(JSON.stringify(results), 50);
       return [
-        `Got a result with multiple return values, expecting a single object from create (${repr})`
+        `Got a non-object result, expecting a single object from create (${repr})`
       ];
     }
 
@@ -24,4 +24,4 @@ const createIsSingle = {
   }
 };
 
-module.exports = createIsSingle;
+module.exports = createIsObject;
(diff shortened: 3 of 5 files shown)


xavdid (changes requested):
So this is great! Couple of things - 

1. Need to tweak the error message to say that we expect a single object
2. Our [docs](https://github.com/zapier/zapier-platform-cli#return-types) say we allow a single-length array as well, but that's [not true](https://github.com/zapier/zapier/blob/develop/developer_cli/api_base.py#L1142). Can you make a quick PR in `cli` (where the docs live) that removes that note from the table? You need to edit `README-source.md` and run `npm run docs` before pushing. 
3. While you're in this code, do you mind glancing over the other checks to make sure the error messages all make sense? There's a typo [here](https://github.com/zapier/zapier-platform-core/blob/master/src/checks/trigger-is-object.js#L25) and we had someone file an [issue](https://github.com/zapier/zapier-platform-cli/issues/249) this morning about a confusing message. I realize this is extending the scope out from under you, but it seems like as good a time as any. 😀

mojowen:
@xavdid here are the error messages:

|  Trigger | Message |
| ------------- | ------------- |
|  *create-is-object*  | `Got a non-object result, expecting a object from create (${repr})` |
| *search-is-array* | `Results must be an array, got: ${typeof results}, (${repr})` |
|  *trigger-is-object*  | `Got a non-object result in the array, expecting an object (${repr})` |
| *trigger-has-unique-ids* | `Got a two or more results with the id of (${doubleId})` |
| *trigger-has-id` | `Got a result missing the "id" property (${repr})` |

src/checks/trigger-is-object.js
```diff
@@ -22,7 +22,9 @@ const triggerIsObject = {
 
     if (nonObjectResult !== undefined) {
       const repr = simpleTruncate(JSON.stringify(nonObjectResult), 50);
-      return [`Got a result missing that was not an object (${repr})`];
+      return [
+        `Got a non-object result in the array, expecting an object (${repr})`
```
xavdid: Since triggers are typically an array of objects (and we want to be consistent with tense), I'd change the end of this error to `expected only objects`)

src/checks/trigger-has-unique-ids.js
```diff
@@ -23,7 +23,7 @@ const triggerHasUniqueIds = {
     });
 
     if (doubleId !== undefined) {
-      return [`Got a two or more results with the id of ${doubleId}`];
+      return [`Got a two or more results with the id of (${doubleId})`];
```
xavdid: minor of minor things, but I'd use double quotes stead of parens. 

```
with the id of "5"
```

reads cleaner than 

```
with the id of (5)
```

since we're being literal. Also, lets be clear about why that's an issue. Tack `Expected all trigger ids to be unique` onto the end of the error.

xavdid (approved):
A couple of tiny things, but these are great! Feel free to merge after the last commit.

src/checks/trigger-has-unique-ids.js
mojowen: Should we change all of the error messages? I used `()` because it was consistent with the others

xavdid: I could go either way. In the other cases, we're putting the whole response object in parens since we're providing general information (rather than saying the object at index 3 is malformed). In this case, we're showing exactly what's wrong (id "5" is not unique), so quotes make more sense to me. 

Either way, super minor. Your call! Merge whatever you like more. If we hate it, ~it's stuck in stone forever~ we'll tweak it further.

mojowen: makes sense - will update to a `"`

PR #68, opened 2018-02-26
Improve error message formatting

nickthetait:
fixes zapier-platform-cli/#249


xavdid:
@nickthetait thanks for fixing that! As luck would have it, we just fixed that in #67. We appreciate you taking the time to make the CLI better though!

PR #69, opened 2018-02-28
bump to 6.10.3

xavdid:
This should be totally safe, just wanted to double check


xavdid:
note that core only ever runs "officially" on `LAMBDA_VERSION`, but these test failures may indicate an incompatibility on node 7+. I can take a look

eliangcs (commented):
Yeah, we can fix Node 7+ in another PR.

eliangcs (approved):

xavdid:
I'll probably fix it here so we don't merge always-failing tests into master. If it's anything complicated, i'll ping you again!

xavdid:
I have no idea what exactly changed between 6 & 7 that made this break, but we're doing a lot of fairly arcane things with promises (injecting bluebird over native promise and then modifying it on our own further). Anyway, fixed in https://github.com/zapier/zapier-platform-core/pull/69/commits/d28fda6f968d334000809e5401e16773acae0811. The answer was a small tweak that *shouldn't* affect anything else, but also this is a lot of magic so who knows.

PR #70, opened 2018-02-28
read from .env by default in inject()

xavdid:
Bumped the dotenv dependency, hence the path changes. Fixes [PDE-76](https://zapierorg.atlassian.net/browse/PDE-76).


test/tools/environment.js
```diff
@@ -1,30 +1,55 @@
...
+      '.env': 'PIZZA="Blackjack"\n',
       secrets: 'SECRET=very_secret_thing'
     });
-  });
 
-  it('should parse a config', done => {
     tools.env.inject();
     process.env.PIZZA.should.equal('Blackjack');
+    should.not.exist(process.env.CITY);
+    done();
```
eliangcs: The tests are synchronous code, so I guess we don't need `done()` here. So are the other test cases in this file.

test/tools/environment.js
```diff
@@ -1,30 +1,55 @@
-require('should');
+const should = require('should');
 const mock = require('mock-fs');
 
 const tools = require('../../src/tools/exported');
-const fake_file = 'CITY="boulder"\nNAME="david"\nPIZZA="Blackjack"\n';
+let env;
 
-describe('read env', () => {
+describe.only('read env', () => {
```
eliangcs: Should we remove `.only` here? Seems mocha [only runs the tests in `test/tools/environment.js`](https://travis-ci.org/zapier/zapier-platform-core/builds/347054760#L1170) because of this.

src/tools/environment.js
```diff
@@ -31,10 +31,29 @@ const cleanEnvironment = () => {
   }
 };
 
+const localFilepath = filename => {
+  return path.join(process.cwd(), filename || '');
```
eliangcs, line 37: Not sure if it's valid use case: Could a user inject an environment file using an absolute path? For example:

```js
const zapier = require('zapier-platform-core');
zapier.tools.env.inject('/absolute/path/to/environment/file');
```

If so, we may want to adjust a bit here because `path.join` doesn't seem to handle absolute paths:

```
> process.cwd()
'/Users/eliang/Projects/zapier-platform/core'
> path.join(process.cwd(), '/root')
'/Users/eliang/Projects/zapier-platform/core/root'
```

src/tools/environment.js
```diff
@@ -31,10 +31,29 @@ const cleanEnvironment = () => {
...
+  if (result.error) {
+    // backwards compatibility
+    result = dotenv.load({ path: localFilepath('.environment') });
+    if (result.parsed) {
+      console.log(
+        [
+          '\nWARNING: `.environment` files will no longer be read by default in the next major version.',
+          'Either rename your file to `.env` or explicitly call this function with a filename:',
+          '\n    zapier.tools.env.inject(".environment");\n\n'
+        ].join('\n')
```
eliangcs, line 55: We may want to hide this when running tests.

test/tools/environment.js
xavdid: oh good catch. that was just for me, since I didn't want to run the whole suite every time. Will remove.

src/tools/environment.js
xavdid: they couldn't, but they also couldn't before. In [master](https://github.com/zapier/zapier-platform-core/blob/master/src/tools/environment.js#L36), we resolved their filename to the local directory anyway. We could expand that, but that seems like a more disruptive change (instead of renaming, they have to pass the whole path in).

xavdid:
Cleaned up in fd59bb7!

eliangcs (approved):

PR #71, opened 2018-02-28
remove marked dep entirely

xavdid:
The core version of [PDE-96](https://zapierorg.atlassian.net/browse/PDE-96). Fixes insecure dependency. Thinking about it though, we don't really need markdown-toc in this repo at all. It was just there to copy the readme from `docs`. Since we don't use this readme for anything anyway, a full removal made more sense.


eliangcs (approved):
Thanks for cleaning it up!

PR #72, opened 2018-03-09
[MAJOR] Remove bundle.environment

xavdid:
It's been deprecated since the first release, so it's a good a time as any to clean it out and make everything a little more consistent. 

**It should be noted** that I haven't been able to test this directly. It turns out it's a bear to build a zip with a different version of `core` in it (probably a good thing).

Anyway, the reason I think this is fine is because I've got an app that's been running an app in production that uses an environment variable but that value isn't present in the bundle:

```
% zapier env 1.0.0
The env of your "Steam" listed below.

┌─────────┬───────────────┬──────────────────────────────────┐
│ Version │ Key           │ Value                            │
├─────────┼───────────────┼──────────────────────────────────┤
│ 1.0.0   │ STEAM_API_KEY │ 8D5A1E0...                       │
└─────────┴───────────────┴──────────────────────────────────┘
```

``` % zapier logs
│     Log       │ { inputDataRaw: {},                                  │
│               │   meta:                                              │
│               │    { frontend: false,                                │
│               │      standard_poll: true,                            │
│               │      prefill: false,                                 │
│               │      test_poll: false,                               │
│               │      limit: -1,                                      │
│               │      hydrate: true,                                  │
│               │      first_poll: false,                              │
│               │      page: 0 },                                      │
│               │   authData: { steamid: ':censored:15:d607b82756:' }, │
│               │   inputData: {},                                     │
│               │   environment: {} }
```

the http logs are free of errors, which would only be the case if the api key was correctly set in the environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,12 +11,11 @@
     "preversion": "git pull && npm test",
     "version": "node bin/bump-dependencies.js && git add package.json",
     "postversion": "git push && git push --tags",
-    "test":
-      "node_modules/.bin/eslint src test && node_modules/.bin/mocha -t 5000 --recursive test",
-    "plain-test": "node_modules/.bin/mocha -t 3000 --recursive test",
-    "integration-test":
-      "node_modules/mocha/bin/mocha -t 10000 integration-test",
-    "lint": "node_modules/.bin/eslint src test",
+    "test": "mocha -t 5000 --recursive test",
+    "posttest": "npm run lint",
+    "plain-test": "mocha -t 3000 --recursive test",
+    "integration-test": "mocha -t 10000 integration-test",
+    "lint": "eslint src test",
     "build": "bin/build.sh local.bundle.zip",
     "upload": "bin/upload-lambda.js local.bundle.zip",
     "deploy":
--- a/src/tools/environment.js
+++ b/src/tools/environment.js
@@ -11,12 +11,8 @@ const { IS_TESTING } = require('../constants');
 // Copy bundle environment into process.env, and vice versa,
 // for convenience and compatibility with native environment vars.
 const applyEnvironment = event => {
-  event = ensurePath(event, 'bundle.environment'); // deprecated
-  _.extend(
-    process.env,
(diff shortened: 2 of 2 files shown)


xavdid:
Ok! I've confirmed this works. 

```
│     Log       │ { inputDataRaw: {},                                  │
│               │   meta:                                              │
│               │    { frontend: false,                                │
│               │      standard_poll: true,                            │
│               │      prefill: false,                                 │
│               │      test_poll: false,                               │
│               │      limit: -1,                                      │
│               │      hydrate: true,                                  │
│               │      first_poll: false,                              │
│               │      page: 0 },                                      │
│               │   authData: { steamid: ':censored:15:d607b82756:' }, │
│               │   inputData: {} }                                    │
│     Version   │ 1.0.0                                                │
│     Step      │ 740e12ab-cdee-4946-b300-6c9c09cfa241                 │
│     Timestamp │ 2018-03-09T15:46:23-06:00
```

eliangcs (approved):
I'm giving a 👍in advance, but I don't have historical context about this change. Can you write a little bit more about what `event.bundle.environment` was and why it was there?

xavdid:
this actually predates me too! @bryanhelmig might have context. The initial public commit had it [marked as deprecated](https://github.com/zapier/zapier-platform-core/blob/c298957d4d69a02c7d2d6ce7dfe2d005d8c62b3e/src/tools/environment.js#L12). It seems like they were originally going to pipe the env into the bundle but instead went with it as a proper env variable. 

Either way, i'm pretty sure it's vestigial now.

eliangcs:
@xavdid I was doing a sanity test with our [github example app](https://github.com/zapier/zapier-platform-example-app-github). This PR seem to break the authentication. When I tried to create an account my Zapier local environment, this error popped up:

screen shot 2018-05-07 at 16 27 29

After reverting this PR, the issue was gone. I'm not sure how this is the case, but apparently somewhere is still using `bundle.environment`. Maybe we should we revert?

PR #73, opened 2018-03-16
[MAJOR] resource collisions cause errors

xavdid:
minor update to previous work, collisions are errors.

### DIFF
--- a/src/tools/schema.js
+++ b/src/tools/schema.js
@@ -115,17 +115,13 @@ const compileApp = appRaw => {
   });
 
   if (problemKeys.length) {
-    // TODO - DB: throw an error instead of logging
-    console.log(
-      '\nWARNING! The following key(s) conflict with those created by a resource:\n'
-    );
-    console.log(
+    const message = [
+      'The following key(s) conflict with those created by a resource:\n',
       problemKeys.map(k => `* ${k}`).join('\n'),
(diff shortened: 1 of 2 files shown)


src/tools/schema.js
```diff
@@ -115,17 +115,13 @@ const compileApp = appRaw => {
...
+    const message = [
+      'The following key(s) conflict with those created by a resource:\n',
       problemKeys.map(k => `* ${k}`).join('\n'),
-      `\n\nEdit the standalone object${
-        problemKeys.length > 1 ? 's' : ''
-      } to resolve`,
-      '!! In the next major version, this will throw an error\n'
-    );
+      '\n\nEdit the standalone object(s) to resolve'
+    ].join('');
```
eliangcs: > Edit the standalone object(s) to resolve

Minor, but the message seems a bit unclear to me. Maybe we can change it to "Rename the key(s) to resolve.".

eliangcs (approved):
Pulled and tested. Works as expected! I only have a suggestion for the error message.

src/tools/schema.js
xavdid: Sure. I just wanted to stress that you modify the standalone objects rather than the resources. Will fix tomorrow, merge eventually!

PR #75, opened 2018-03-29
Raise Exception for Bad Requests When Stashing A File

### LINKED ISSUE #74: stashFile Should Handle Failed File with Better Error Messaging
In cases where stashFile is passed a request with the `raw: true`, it will sometimes evaluate the following cases as true:

https://github.com/zapier/zapier-platform-core/blob/1153a059f6b66768dfcd9e11f0a96475cc106e46/src/tools/create-file-stasher.js#L115-L119

Which leads stashFile to call `response.content` - which raises the following error:

https://github.com/zapier/zapier-platform-core/blob/1153a059f6b66768dfcd9e11f0a96475cc106e46/src/http-middlewares/after/prepare-response.js#L40-L44

This error doesn't help explain the underlying reason for the failure - the unsuccessful request for a file.

mojowen:
## Description
Checks the status of a request for a file when attempting to stash said file

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :bug: Bug fix  |

## Related Issue

Fixes #74

## Before / Current Behavior
Currently no checks are made when buffering a response into the file stasher - leading to `response.content` being called on a buffer, which raises an exception

## After / New Behavior
Will throw an exception if a non 200 status is returned when attempting to retrieve a file

### DIFF
--- a/src/tools/create-file-stasher.js
+++ b/src/tools/create-file-stasher.js
@@ -117,7 +117,6 @@ const createFileStasher = input => {
                 } else {
                   newBufferStringStream = response.content;
                 }
-
                 knownLength =
                   knownLength || response.getHeader('content-length');
                 const cd = response.getHeader('content-disposition');
@@ -141,8 +140,8 @@ const createFileStasher = input => {
                 fileContentType
               );
             };
-
             if (isStreamed) {
+              maybeResponse.throwForStatus();
               return maybeResponse.buffer().then(parseFinalResponse);
             } else {
               return parseFinalResponse(maybeResponse);
--- a/test/tools/file-stasher.js
+++ b/test/tools/file-stasher.js
@@ -4,6 +4,7 @@ const fs = require('fs');
 const path = require('path');
 
 const should = require('should');
+const nock = require('nock');
 
 const mocky = require('./mocky');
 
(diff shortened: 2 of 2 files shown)


test/tools/file-stasher.js
```diff
@@ -115,6 +116,27 @@ describe('file upload', () => {
       .catch(done);
   });
 
+  it('should throwForStatus if bad status', done => {
+    mocky.mockRpcCall(mocky.fakeSignedPostData);
```
mojowen, line 120: this is cruft - will delete

src/tools/create-file-stasher.js
```diff
@@ -141,8 +140,8 @@ const createFileStasher = input => {
                 fileContentType
               );
             };
-
             if (isStreamed) {
+              maybeResponse.throwForStatus();
```
mojowen, line 144: may be smart to typecheck here to make sure this object is a request - although maybe it's a safe assumption as we're calling `.buffer` on it anyway?

xavdid (approved):
looks good, great catch!

test/tools/file-stasher.js
mojowen: whoops - nope this is required. Readding

PR #76, opened 2018-04-04
z.cursor store

xavdid:
Covers [PDE-142](https://zapierorg.atlassian.net/browse/PDE-142).

This is being tested with [this trigger](https://gist.github.com/xavdid/b3a1c69ecc410593009dde4f421acd26)

**Note**: https://github.com/zapier/zapier/pull/16646 needs to be deployed before this can be released

See linked `z/z` PR for testing info. 

CI is failing because it's still pointing at my local ngrok for testing reasons. I'll remove that before this is merged.

### DIFF
--- a/src/tools/create-app-tester.js
+++ b/src/tools/create-app-tester.js
@@ -3,6 +3,22 @@
 const createLambdaHandler = require('./create-lambda-handler');
 const resolveMethodPath = require('./resolve-method-path');
 const ZapierPromise = require('./promise');
+const { get, last } = require('lodash');
+const { genId } = require('./data');
+
+// this is (annoyingly) mirrored in cli/api_base, so that test functions only
+// have a storeKey when canPaginate is true. otherwise, a test would work but a
+// poll on site would fail. this is only used in test handlers
+const shouldPaginate = (appRaw, method) => {
+  const methodParts = method.split('.');
+  if (methodParts[0] !== 'triggers' || last(methodParts) !== 'perform') {
+    return false;
+  }
+
+  methodParts.pop();
+
+  return get(appRaw, `${methodParts.join('.')}.canPaginate`);
+};
 
 // Convert a app handler to promise for convenience.
 const promisifyHandler = handler => {
@@ -33,7 +49,7 @@ const createAppTester = appRaw => {
       command: 'execute',
       method,
       bundle,
-      storeKey: 'testKey'
+      storeKey: shouldPaginate(appRaw, method) ? `testKey-${genId()}` : null
     };
 
     if (process.env.LOG_TO_STDOUT) {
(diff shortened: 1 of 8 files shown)


src/tools/create-app-tester.js
```diff
@@ -3,6 +3,22 @@
...
 const resolveMethodPath = require('./resolve-method-path');
 const ZapierPromise = require('./promise');
+const { get, last } = require('lodash');
+const { genId } = require('./data');
+
+// this is (annoyingly) mirrored in cli/api_base, so that test functions only
+// have a storeKey when canPaginate is true. otherwise, a test would work but a
+// poll on site would fail. this is only used in test handlers
+const shouldPaginate = (appRaw, method) => {
+  const methodParts = method.split('.');
```
BrunoBernardino: Could this more easily understood with `method.startsWith('triggers') && method.endsWith('perform')`, then not needing `last` from `lodash`?

Feel free to ignore as you need the `methodParts` below anyway.

src/tools/create-storekey-tool.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const _ = require('lodash');
+const ZapierPromise = require('./promise');
+
+// Similar API to JSON built in but catches errors with nicer tracebacks.
```
BrunoBernardino: I believe this comment came from copy-🍝 .

test/tools/rpc-client.js
```diff
@@ -29,4 +29,26 @@ describe('rpc client', () => {
       })
       .catch(done);
   });
+
+  it('should set a cursor key', done => {
```
BrunoBernardino: I'd love to see a test getting a cursor that doesn't exist, and setting an invalid store key (like `undefined` or `null` or a gigantic string), just so we can be sure failures are failing 😄

BrunoBernardino (approved):
Great stuff! Love the simplicity of this API and the tests.

I didn't pull it down as it's a bit convoluted to test, and you've got test coverage.

I've got a few questions/#suggestions, but not really blocking.

src/tools/create-app-tester.js
xavdid: we need to split it anyway, but I don't like checking on the string rather than the array. good call!

test/tools/rpc-client.js
xavdid: yeah! getting a cursor that hasn't been set is fine and returns `null` (just like redis does). A store key missing is an error, and that behavior has a unit test on the server. I can go ahead and add something there though!

xavdid: I'm having trouble testing this because since we're not actually sending any calls, nothing actually fails unless I explicitly fail it, which is then just the same. Given that there's actually a [server test](https://github.com/zapier/zapier/pull/16646/files#diff-c7aa4a0aa5350b7313803bc516e8dce5R100), i'll skip here for now.

PR #77, opened 2018-04-19
Better log censoring

eliangcs:
Addresses #4 and zapier/zapier#16839.

So besides the original censoring on `authData`, this PR adds another step censoring based on the ad-hoc field names, similar to how [Sentry](https://github.com/getsentry/sentry/blob/8.10.0/src/sentry/constants.py#L166-L175) and [Django](https://docs.djangoproject.com/en/2.0/ref/settings/#debug) do it.


BrunoBernardino (approved):
Love it! Awesome tests, great clean solution!

I didn't pull it down because of the great tests, and because it looks straightforward reading!

PR #78, opened 2018-04-23
PoC: recurseCleanFuncs: option to preserve code

FokkeZB:
Quick proof of concept that we could generate a `definition.json` which includes the actual function code. 

An example of the output for https://github.com/zapier/zapier-platform-example-app-onedrive:

https://gist.github.com/FokkeZB/c339a9cdc1220e260eca43148fbf0da9

Note: I replaced `path` since it didn't seem to be used anywhere?

### DIFF
--- a/src/tools/cleaner.js
+++ b/src/tools/cleaner.js
@@ -6,23 +6,25 @@ const isPlainObj = require('./data').isPlainObj;
 const recurseReplace = require('./data').recurseReplace;
 const flattenPaths = require('./data').flattenPaths;
 
-const recurseCleanFuncs = (obj, path) => {
+const recurseCleanFuncs = (obj, toString = false) => {
   // mainly turn functions into $func${arity}${arguments}$
-  path = path || [];
   if (typeof obj == 'function') {
+    if (toString) {
+      return obj.toString();
+    }
     const usesArguments =
       obj.toString().indexOf('arguments') !== -1 ? 't' : 'f';
     // TODO: could optimize $func$0$f$ as "pure" and just render them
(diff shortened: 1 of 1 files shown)


bryanhelmig:
This is interesting @FokkeZB!

I can think of two alternative ways to do something like this:

### Source Embedded in JS

https://github.com/zapier/zapier-platform-core/pull/80 implements a small portion of this:

```js
const App = {
  resources: {
    contact: {
      key: 'contact',
      noun: 'Contact',
      list: {
        display: {},
        operation: {
          perform: {
            { source: 'return [{ id: 45678 }];' }
          }
        }
      }
    }
  }
};
```

Then we could recurse into that and do `new Function('z', 'bundle', source)`.

@xavdid talked a bit about something like this in his https://zapier.quip.com/wXRfAQ468YHO which I think is pretty cool! I go on at length here https://zapier.quip.com/Gl6cAKKbsUz9.

### Standalone File Requiring

```js
const App = {
  resources: {
    contact: {
      key: 'contact',
      noun: 'Contact',
      list: {
        display: {},
        operation: {
          perform: {
            { require: 'src/contact/list.bd8e6f.js' }
          }
        }
      }
    }
  }
};
```

And then you might do something like `loadOrRequire(file)` and that "file" could be passed in as an argument to the Lambda function for overriding.

`src/contact/list.bd8e6f.js` should just `module.exports = (z, bundle) => {};` something.

FokkeZB:
Thanks @bryanhelmig - I've also now borrowed from this in https://zapier.quip.com/2SPZAk4TxbpT#INMACAO2nFM

PR #79, opened 2018-04-27
Add support for multiple resource actions.

bryanhelmig:
Picks up where https://github.com/zapier/zapier-platform-schema/pull/43 left off.

Still needs updated tests.

### DIFF
--- a/src/tools/schema.js
+++ b/src/tools/schema.js
@@ -8,23 +8,23 @@ const zapierSchema = require('zapier-platform-schema');
 // method is {display: {}, operation: {}}
 // methodMap is {key: {display: {}, operation: {}}}
 const walkMethods = (resource, methodOrMethodMap) => {
-  const _default = [];
-  if (!methodOrMethodMap) {
-    return _default;
-  }
+  methodOrMethodMap = methodOrMethodMap || {};
+
   if (methodOrMethodMap.operation) {
     const method = dataTools.deepCopy(methodOrMethodMap);
     method.key = resource.key;
     return [method];
-  } else if (Object.keys(methodOrMethodMap).length) {
+  }
+
+  if (Object.keys(methodOrMethodMap).length) {
     return Object.keys(methodOrMethodMap).map(key => {
       const _method = dataTools.deepCopy(methodOrMethodMap[key]);
       _method.key = `${resource.key}_${key}`;
       return _method;
     });
-  } else {
-    return _default;
   }
+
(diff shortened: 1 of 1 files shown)


src/tools/schema.js
```diff
@@ -5,6 +5,28 @@ const cleaner = require('./cleaner');
...
+const walkMethods = (resource, methodOrMethodMap) => {
+  methodOrMethodMap = methodOrMethodMap || {};
+
+  if (methodOrMethodMap.operation) {
+    const method = dataTools.deepCopy(methodOrMethodMap);
+    method.key = resource.key;
+    return [method];
+  }
+
+  if (Object.keys(methodOrMethodMap).length) {
```
xavdid, line 19: Would probably use `_.isEmpty(methodorMethodMap)` instead

xavdid: and actually, don't even really need this check here, since `Object.keys({}).map(...)` returns `[]` anyway. Though it might be worth keeping from a clarity standpoint.

src/tools/schema.js
```diff
@@ -13,43 +35,50 @@ const convertResourceDos = appRaw => {
...
+    });
+
+    walkMethods(resource, resource.misc).forEach(misc => {
+      misc.key = `${resource.key}Misc`;
+      misc.noun = resource.noun;
+      misc.operation.resource = resource.key;
+      creates[misc.key] = misc;
+    });
+
+    // TODO: ergh... this won't work with multiple searches/creates
```
xavdid, line 75: I think we could enforce some sort of naming pattern with the keys to automatically tie them together. It would be a breaking change, but not too bad of one. In that case, we'd iterate the searches and see if there's a create with a matching key. totally doable!

xavdid (commented):
this is cool! I think we're in a pretty good spot. as always, we'll also want some tests. 😀

src/tools/schema.js
bryanhelmig: Oh, good points!

src/tools/schema.js
bryanhelmig: Could tie them together if one search and one create, and but if more than 1 for either, fall back to specific naming patterns.

xavdid:
@bryanhelmig given there hasn't been movement here (or in https://github.com/zapier/zapier-platform-schema/pull/43) I think i'm going to close it unless you've got objections.

PR #80, opened 2018-05-01
Self enclosed dynamic schema.

bryanhelmig:
Handle a new case for `perform` functionality:

* _old_ `request` which is a hardcoded request object with embedded, lazy `{{}}` evaluation. EG: https://github.com/zapier/zapier-platform-schema/blob/master/docs/build/schema.md#requestschema
* _old_ `func` which is hardcoded functions from an on disk JS AST / zip. EG: https://github.com/zapier/zapier-platform-schema/blob/master/docs/build/schema.md#functionschema
* _new_ `source` which is `perform = {source: 'return [{id: 123}]'}` which we turn into `Function('z', 'bundle', perform.source)`. EG: https://github.com/zapier/zapier-platform-schema/pull/45 (an extension of above `func`).
* _new_ `require` which is `perform = {requuie: 'some/path/to/file.js'}` which we turn into `require(perform.require)`. EG: https://github.com/zapier/zapier-platform-schema/pull/45 (an extension of above `func`).

This is the first step in supporting CLI apps that could be "zipless" -- IE: just hand a generic core runner any self-enclosed schema and we can run it. This should be the first step on a path to no-build basic apps with JS support.

- [x] support `event.appRawOverride` in `loadApp(appRawOrPath)`.
- ~~[ ] explore the merging of app definitions (IE: `merge(appRaw, event.appRawOverride)`)~~  (maybe later)

### DIFF
--- a/integration-test/integration-test.js
+++ b/integration-test/integration-test.js
@@ -114,6 +114,33 @@ const doTest = runner => {
         .catch(done);
     });
 
+    it('should handle appRawOverride', done => {
+      const event = {
+        command: 'execute',
+        method: 'triggers.fooList.operation.perform',
+        appRawOverride: {
+          resources: {
+            foo: {
+              key: 'foo',
+              noun: 'Foo',
+              list: {
+                display: {},
+                operation: {
+                  perform: { source: 'return [{id: 45678}]' }
+                }
+              }
+            }
+          }
+        }
+      };
+      runner(event)
+        .then(response => {
+          response.results.should.deepEqual([{ id: 45678 }]);
+          done();
+        })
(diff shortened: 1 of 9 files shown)


bryanhelmig:
Ah, just realized we couldn't provide `callback` as the third argument without screwing up the `execute()` which will think it won't return a promise. Maybe could do `code.indexOf('callback') !== -1` but yuck.

bryanhelmig:
I think this is ready for review!

BrunoBernardino (approved):
Looks pretty slick! I didn't pull this down or test, for brevity sake, but can do next week!

PR #81, opened 2018-05-08
Build boilerplate and upload to S3

BrunoBernardino:
The bucket still needs to be created and I need to fully test the whole thing together.

Related to https://github.com/zapier/zapier/pull/17303 but it doesn't need it shipped in order to ship this one (only the bucket to be created).


package.json
```diff
@@ -10,15 +10,18 @@
   "scripts": {
     "preversion": "git pull && npm test",
     "version": "node bin/bump-dependencies.js && git add package.json",
-    "postversion": "git push && git push --tags",
+    "postversion": "git push && git push --tags && npm run deploy-boilerplate",
```
eliangcs: We're using Travis to help us release `cli/core/schema` packages to npm. Maybe we want to do that too with the boilerplate? IE: Whenever there's a new tag on GitHub, Travis builds a new boilerplate and upload it to S3.

eliangcs (approved):
Pulled and confirmed the boilerplate is built correctly. I'm not sure if `core` is the best place to put boilerplate as the boilerplate doesn't relate to `core` that much. One good thing I can think of is that it's easy to manage boilerplate uploads along with CLI releases, so I'm fine with this solution.

I have a minor suggestion of moving `deploy-boilerplate` to Travis. Don't have to do it in this PR, but I think that's what we're going for eventually.

package.json
BrunoBernardino: Oh, I didn't realize we'd changed that process. Looking at https://zapier.quip.com/bns4AxqwaMIm I'd still expect us to have to run `npm version`  (which would trigger this and the travis build).

Am I missing something? Otherwise, please let me know what should I change/where to get that working. I do want it to be as automatic as possible.

BrunoBernardino:
Thanks @eliangcs !

The reason why I put it in `core` was because it uses the `zapierwrapper.js` (which is already here), and it had the dev dependencies we needed (AWS).

It also could be in `cli` as it requires `core`. 🤷‍♂️

I started by trying to put it in `schema` (because it exports the schema), but I'd have to include `core` or download the `zapierwrapper.js` file from it, so I moved it here instead.

package.json
eliangcs: The process is pretty much the same. We still need to run `npm version` locally. It just that `npm version` now does less -- it bumps version, adds a git tag, pushes to GitHub, and that's it. The rest of the deployment is done by Travis. So my point is, I think it's better to keep `npm version` doing as little as possible.

Travis supports [S3 deployment](https://docs.travis-ci.com/user/deployment/s3/). It's likely we can add another `provider` in the `deploy` section of `.travis.yml` to do the boilerplate deployment. This is not a blocker -- we can definitely leave it to another PR.

bryanhelmig (approved):
Agree with @eliangcs's comments, but this looks good! We can work on getting that S3 bucket up for you.

package.json
jamescarr: Look to the playbook for an example where we do publish to S3 from Travis. :-)

jamescarr: See https://github.com/zapier/playbook/blob/master/.travis.yml#L8

.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 *.zip
 *.log
 .zapier-*
+.DS_Store
```
xavdid, line 5: fine to have here, but it's worth adding to your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) (along with `.env` and stuff you'll never want in source control on purpose)

bin/build-boilerplate.sh
```diff
@@ -0,0 +1,45 @@
...
+# This allows us to avoid duplicating files in git
+FILES_TO_COPY='zapierwrapper.js'
+
+if [ -f $FILE ]
+then
+   rm -f $FILE
+fi
+
+# Copy files
+cp include/$FILES_TO_COPY boilerplate/
```
xavdid: I'm rusty on bash, but I believe it's best practice to wrap expansions in paths in quotes: `cp "include/$FILES_TO_COPY" boilerplate/`. That way, if we add more files to copy later, we won't worry about weird path expansions other places in the file

xavdid (approved):
Looks good! 

My only major comment is that it might be worth doing `build-boilerplate.sh` in node rather than bash. Working on the json itself seems a little more stable than doing a text replace on it and the less bash knowledge we need to maintain a node project, the better (imo). 

that said, it's short and well commented, so it's not a huge deal. `cd` and `npm install` might be tough using a node script. If we don't want to do non-bash entirely, I'm fine leaving as is. We could also meet halfway and drop a little node cli function do to the renaming and stuff (that we call from this script), but that might be overkill.

BrunoBernardino:
Thanks @xavdid ! I did start with a node script, but it was getting a bit complicated with many file writes and process execs, so bash simplified it, even if, like you said, isn't ideal for the json replacements. I'd definitely prefer to add a small node script if we need to do more replacements.

BrunoBernardino:
@jamescarr can you please check if this travis deploy file is OK (and potentially add the settings to Travis)?

jamescarr:
@BrunoBernardino yep and I'll add the credentials in too

BrunoBernardino:
Thank you so much @jamescarr ! Merging!

PR #86, opened 2018-05-11
Ability to run legacy scripting in embedded JS

eliangcs:
This PR makes `z.legacyScripting.run()` available so we can run legacy scripting easily in the embedded JS. The usage can be found in the [newly-added test](https://github.com/zapier/zapier-platform-core/pull/86/files#diff-0f26e4e23314fbbe48c885143da18751R29).

We need to ship https://github.com/zapier/zapier-platform-legacy-scripting-runner/pull/3 first to the make the test on Travis pass. To test locally, you need to clone `zapier-platform-scripting-legacy-runner` and run `npm link` there. And then in your local `core` repo, run `npm install && npm link zapier-platform-legacy-scripting-runner`.


bin/build-boilerplate.sh
```diff
@@ -36,8 +36,17 @@ sed -i '' -e "s/$FIND_DEF_STRING/$REPLACE_DEF_STRING/g" "$CONTENTS_DIR/definitio
...
+  cp -R ../src/* node_modules/zapier-platform-core/src/
+  cp -R ../node_modules/zapier-platform-legacy-scripting-runner/* node_modules/zapier-platform-legacy-scripting-runner/
+  rm -rf node_modules/zapier-platform-core/node_modules
+  rm -rf node_modules/zapier-platform-legacy-scripting-runner/node_modules
+  rm -rf node_modules/zapier-platform-schema/node_modules
+fi
+
 # Build the zip file!
-zip -R ../$FILE '*.js' '*.json'
+zip -R ../$FILE '*.js' '*.json' '*.node'
```
eliangcs: `*.node` is to include the binary files necessary for package `deasync`.

BrunoBernardino: How big is the new `zip` file? No problem with it?

BrunoBernardino (approved):
Awesome stuff, and with tests! 👍

bin/build-boilerplate.sh
eliangcs: The new zip file is about 4.0 MB. One thing we can do to make it smaller is to exclude the binaries that are for Windows or Node version < 6. I'll push another patch for that. Thanks!

eliangcs: Addressed in cb6852f2c0dcc.

test/legacy-scripting/legacy-scripting.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
```
BrunoBernardino, line 1: Much nicer structure now!

BrunoBernardino (approved):
Looking great! Didn't pull as tests are passing. 👍 

Love the new tests organization for this.

boilerplate/package.json
```diff
@@ -11,6 +11,9 @@
     "npm": ">=3.10.10"
   },
   "dependencies": {
-    "zapier-platform-core": "CORE_PLATFORM_VERSION"
+    "async": "2.6.0",
+    "moment-timezone": "0.5.16",
+    "zapier-platform-core": "CORE_PLATFORM_VERSION",
+    "zapier-platform-legacy-scripting-runner": "1.2.0"
```
bcooksey, line 17: Would it be possible to have legacy-scripting-runner, `moment-timezone`, and `async` be part of the default lambda app we upload, vs being in core itself @eliangcs @BrunoBernardino?

BrunoBernardino: I'm not sure I understand the question. This will be the default lambda app we upload, it's not `core` itself (notice it's `boilerplate/package.json`, not `package.json`).

Sorry if I'm misunderstanding the question here!

eliangcs: @bcooksey not sure if I understand it correctly. Correct me if I misunderstood.

This `boilerplate/package.json` file **is** the package.json for the default lambda app we upload. We call it this default runtime `boilerplate.zip`. The only reason why it's in `core` is that we build `boilerplate.zip` in `core` ([PR](https://github.com/zapier/zapier-platform-core/pull/81)). The `core` itself doesn't require `legacy-scripting-runner`, `moment-timezone`, etc.

bcooksey: Sorry, I commented in the wrong spot, but if you look at [the diff for package.json](https://github.com/zapier/zapier-platform-core/pull/86/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) it looks like all those leaked into core itself, right?

This may deserve a real-time discussion, but I'm wondering if there is a way to not pollute core with code from a converted app. Is there a way to separate concerns here?

BrunoBernardino: Ah, right. It's a part of `devDependencies` and necessary for the `boilerplate`, as `core` builds `boilerplate/X.zip` on release (it doesn't get into apps). Does that make sense?

eliangcs: @bcooksey another reason why they're in `devDependencies` is that we have legacy scripting tests in `core` [here](https://github.com/zapier/zapier-platform-core/blob/649866eafeaaea0edbf196a40425bbcf8c1f7bcf/test/legacy-scripting/legacy-scripting.js). I actually thought about moving all of these (the tests, `moment-timezone`, `async`, etc) to `legacy-scripting-runner` to free `core` from knowing the existence of `legacy-scripting-runner`. But that didn't work because we need to [create a z-app](https://github.com/zapier/zapier-platform-core/blob/649866eafeaaea0edbf196a40425bbcf8c1f7bcf/test/legacy-scripting/legacy-scripting.js#L34) to run those tests, which can't be done in `legacy-scripting-runner`. It's a bit hacky, but I wish I had a better solution.

bcooksey: Got it, makes sense!

PR #88, opened 2018-05-16
Added a number of additional typescript definitions

toddbluhm:
This is by no means a complete typescript definition file, but I created as much as I needed for my zapier app. I figure I would at least throw up this PR if the zapier team or anyone else wants to use this as starting point to finish off the rest of the definitions.

I don't expect this to be merged as is, but its a good start on completing the typescript definitions.

A lot of the definitions were taken straight from the zapier schemas here: https://zapier.github.io/zapier-platform-schema/build/schema.html

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,6 +8,7 @@ import { Agent } from 'http';
 
 // The EXPORTED OBJECT
 export const version: string;
+export const Promise: PromiseLike<any>;
 export const tools: { env: { inject: (filename?: string) => void } };
 export const createAppTester: (
   appRaw: object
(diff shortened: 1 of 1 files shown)


xavdid:
@toddbluhm this is awesome! I really appreciate you taking the time to put this stuff together. 

Since the schema is so large and changes periodically, I've been messing with a script that will convert the schema definition into typings (using [json-schema-to-typescript](https://github.com/bcherny/json-schema-to-typescript)). There have been some hiccups which is why it's not in a workable state yet. 

I think for now i'm going to leave this here and point to it as a great starting place if devs want to use some of the schemas. I don't see maintaining it manually as a viable long-term solution though (especially when I'm the only typescript person on the team 😁)

toddbluhm:
Closing this out as its probably well out of date by now.

xavdid:
@toddbluhm You may be right. In any case, I appreciate the work you put in here. This is something that's definitely still on my backlog and kick the tires on the tool I mentioned above every once in a while just to see if we're there yet. I'm not sure when I'll have the time on this, but it'll happen!

PR #89, opened 2018-05-18
Allow to functions accepting any args besides (z, bundle)

eliangcs:
Currently, the syntax of `{ source: 'return 123;' }` generates a function that accepts `(z, bundle)` as arguments. While this is mostly fine, not all functions in a CLI app take `(z, bundle)`. One example is the middleware functions in `beforeRequest` and `afterResponse`, which take `(request, z, bundle)` as arguments.

This PR fixes it by allowing the app definition to specify `args` as well. For example, `{ source: 'return a + b;', args: ['a', 'b'])`. If `args` isn't provided, we fall back to `(z, bundle)`.

### DIFF
--- a/src/tools/schema-tools.js
+++ b/src/tools/schema-tools.js
@@ -2,10 +2,10 @@
 
 const dataTools = require('./data');
 
-// TODO: bryanhelmig, could support callback third argument but not naively
-const makeFunction = source => {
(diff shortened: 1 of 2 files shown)


BrunoBernardino (approved):
Simple, neat, and with tests! 👍

PR #90, opened 2018-05-18
Fix bug: Middleware isn't compiled

eliangcs:
Fixes the bug where `beforeRequest` and `afterResponse` middlewares don't get compiled from an embedded form like `{ source: 'return request;' }`.


BrunoBernardino (approved):
Nice improvement, with tests!

PR #91, opened 2018-05-23
Fix missed logs

BrunoBernardino:
Makes sure we don't replace the values in certain _key_ keys (ha!).

Partial fix for https://github.com/zapier/zapier/issues/17647 (it has more info about this issue as well).


bryanhelmig (approved):
Good find!

PR #93, opened 2018-06-07
Move boilerplate dependencies update to node

### LINKED ISSUE #92: Boilerplate deployment failed on Travis
The `./bin/build-boilerplate.sh` bash script failed on Traivs: https://travis-ci.org/zapier/zapier-platform-core/jobs/388670692#L1357-L1363 and blocked the deployment. I [removed](https://github.com/zapier/zapier-platform-core/commit/bc4b79210250bf830531fed7c2fa19b5312cccfe) it from `.travis.yml` temporarily to unblock npm release.

Looks like we need to make those `sed` commands Linux compatible. Or we can write the renaming of versions into a Node script.

BrunoBernardino:
Should allow it to run without problems in Travis for the furure.

Fixes #92

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ deploy:
   on:
     tags: true
     node: 6.10.3
+before_deploy:
+- npm run build-boilerplate
 env:
   global:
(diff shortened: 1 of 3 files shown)


eliangcs (approved):
Had a sanity test and it works on local. We'll see if it works on Travis next time we have a new release. Thanks for fixing this!

PR #94, opened 2018-06-12
Bump versions of dependencies and Node.js

kola-er:
Add support for Node.js 8.10.


kola-er:
Until `zapier-platform-schema@7.0.0` is published, the test would keep failing. Waiting on https://github.com/zapier/zapier-platform-schema/pull/48 to be merged and published on `npm`.

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: node_js
 node_js:
 - 6.10.3
 - '7'
```
xavdid: we can remove 7, it'll never be on AWS.

we could also add `'10'` just to see if there's anything bad (eventually) coming down the pipe

package.json
```diff
@@ -36,8 +36,8 @@
     "dotenv": "5.0.1",
     "form-data": "2.2.0",
     "lodash": "4.17.10",
-    "node-fetch": "1.7.1",
-    "zapier-platform-schema": "6.1.0"
+    "node-fetch": "2.1.2",
+    "zapier-platform-schema": "7.0.0"
```
xavdid: this gets bumped automatically on release, so we shouldn't change it manually here.

package.json
```diff
@@ -36,8 +36,8 @@
     "dotenv": "5.0.1",
     "form-data": "2.2.0",
     "lodash": "4.17.10",
-    "node-fetch": "1.7.1",
-    "zapier-platform-schema": "6.1.0"
+    "node-fetch": "2.1.2",
```
xavdid: is this is a safe change to make? Seems like there are some actual [incompatibilities](https://github.com/bitinn/node-fetch/blob/master/UPGRADE-GUIDE.md) here, though I don't know if we're using any of these one way or the other.

xavdid (changes requested):
the only blocker is the schema change, assuming you've checked out that fetch bump

package.json
kola-er: Looking around the [changelog](https://github.com/bitinn/node-fetch/blob/master/CHANGELOG.md), I don't see any 🔥using `2.1.2`. This part of [v2.0.0 general changes](https://github.com/bitinn/node-fetch/blob/master/CHANGELOG.md#general-changes) pushed me to bump it though:

> Enhance: start testing on Node.js v4.x, v6.x, v8.x LTS, as well as v9.x stable

kola-er: To be on a safe side, let's stick with `1.7.1` in this release since it isn't broken running on Node.js `8.10.0`.

xavdid (approved):

PR #96, opened 2018-06-13
Move legacy scripting tests out of core

eliangcs:
Found a way to move legacy scripting tests from `core` to `legacy-scripting-runner` (see https://github.com/zapier/zapier-platform-legacy-scripting-runner/pull/7), so that we can remove the dependencies that are actually for `legacy-scripting-runner` from `core`'s `devDependencies`.


BrunoBernardino (approved):
Great idea!

PR #97, opened 2018-06-13
Whitelist files in package

eliangcs:
Same stuff as https://github.com/zapier/zapier-platform-cli/pull/330.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,10 @@
   "license": "UNLICENSED",
   "main": "index.js",
   "typings": "index.d.ts",
+  "files": [
+      "include/*.js",
+      "src/**/*.js"
+  ],
   "scripts": {
     "preversion": "git pull && npm test",
     "version": "node bin/bump-dependencies.js && git add package.json",
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
   "main": "index.js",
   "typings": "index.d.ts",
   "files": [
-      "include/*.js",
-      "src/**/*.js"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -8,6 +8,10 @@
   "license": "UNLICENSED",
   "main": "index.js",
   "typings": "index.d.ts",
+  "files": [
+      "/include/*.js",
+      "/src/**/*.js"
```
xavdid: Needs the following:

* index.js
* index.d.ts

package.json
```diff
@@ -8,6 +8,10 @@
   "license": "UNLICENSED",
   "main": "index.js",
   "typings": "index.d.ts",
+  "files": [
+      "/include/*.js",
```
xavdid: like `cli`, I think this can be `/include/` and `/src/`, though double check the output of `npm pack` to verify that.

xavdid (changes requested):

package.json
eliangcs: Thanks! Added the two files in c86230c.

NOTE: `index.js` is automatically included as it's specified as `main`, but I added it to `files` anyway just to be more explicit.

package.json
eliangcs: Fixed in c86230c.

eliangcs:
@xavdid updated the PR. Thanks!

xavdid:
cool. didn't explicitly test, but it looks good!

xavdid (approved):

PR #98, opened 2018-06-19
Delegate legacy scripting source compilation to runner

eliangcs:
Lets `legacy-scripting-runner` compile the legacy scripting source (zapier/zapier-platform-legacy-scripting-runner/pull/10), so we don't have to do it here in `core`.


BrunoBernardino (approved):
Awesome stuff!

PR #100, opened 2018-06-20
Bump Node version to 8 for boilerplate

eliangcs:
Bumps versions of Node.js and legacy-scripting-runner for boilerplate.


bin/build-boilerplate.sh
```diff
@@ -47,8 +47,10 @@ zip -R ../$FILE '*.js' '*.json'\
 # Revert copied files
 rm -f $FILES_TO_COPY
 
-# Remove node_modules
-rm -fr node_modules && cd ..
+# Remove generated package-lock.json and node_modules
+rm -f package-lock.json
```
BrunoBernardino, line 51: Ah, nice!

boilerplate/package.json
```diff
@@ -3,17 +3,15 @@
   "version": "1.0.0",
   "description": "A boilerplate that allows running a buildless app.",
   "homepage": "https://zapier.com/",
-  "author": "Zapier <partners@zapier.com>",
+  "author": "Zapier Engineering <engineering@zapier.com>",
```
BrunoBernardino: We usually always ask for people to reach out at partners@zapier.com. Any reason for this change (and should we change in the other packages)?

BrunoBernardino (approved):
Looks good, I've just got a minor question.

boilerplate/package.json
eliangcs: That's a good point. I reverted it back. I made this change only because the author of our CLI packages on npm is engineering@zapier.com.

PR #102, opened 2018-06-29
Rename to ZObject

janaagaard75:
Rename `zObject` to `ZObject` to follow the de facto standard of capitalizing interface names.

PR #103, opened 2018-07-10
(improvement) better `AppTester` typescript bindings

xavdid:
Fixes https://github.com/zapier/zapier-platform-cli/issues/339 and duplicates https://github.com/zapier/zapier-platform-core/pull/102

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,10 +20,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.44",
-        "jsesc": "^2.5.1",
-        "lodash": "^4.2.0",
-        "source-map": "^0.5.0",
-        "trim-right": "^1.0.1"
+        "jsesc": "2.5.1",
+        "lodash": "4.17.10",
+        "source-map": "0.5.7",
+        "trim-right": "1.0.1"
       }
     },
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -20,10 +20,10 @@
...
       "requires": {
         "@babel/types": "7.0.0-beta.44",
-        "jsesc": "^2.5.1",
-        "lodash": "^4.2.0",
-        "source-map": "^0.5.0",
-        "trim-right": "^1.0.1"
+        "jsesc": "2.5.1",
+        "lodash": "4.17.10",
+        "source-map": "0.5.7",
+        "trim-right": "1.0.1"
```
eliangcs, line 26: I keep seeing PRs fighting between carets and no carets in `package-lock.json`. Is it because we're using different versions of npm?

eliangcs (approved):
Didn't test it but this looks good with my limited TypeScript knowledge. I have a question on `package-lock.json` but it's not relevant to this PR.

package-lock.json
xavdid: honestly, I have no idea. It might be because of npm version, and npm setting, or the way that individual dependencies declare _their_ dependencies ¯\\\_(ツ)_/¯

PR #107, opened 2018-08-02
Optimize boilerplate build

eliangcs:
Mainly fixes the following section when we do `./bin/build-boilerplate.sh --debug`:

```
rm -rf node_modules/zapier-platform-core/node_modules
rm -rf node_modules/zapier-platform-legacy-scripting-runner/node_modules
rm -rf node_modules/zapier-platform-schema/node_modules
```

This works fine if all the three packages (`core`, `legacy-scripting-runner`, `schema`) share common dependencies. But if one of them adds a dependency that the others don't rely on, the script generates a broken build.

This PR changes it so that when building a `--debug` boilerplate, we do `npm pack` in `core` and `legacy-scripting-runner` locally. Then we `npm install` those local packages, allowing npm to optimize dependency tree while the code of `core` and `legacy-scripting-runner` is copied from the developer's local machine.

### DIFF
--- a/bin/build-boilerplate.sh
+++ b/bin/build-boilerplate.sh
@@ -24,25 +24,21 @@ NEW_VERSION="$(node -p "require('./package.json').version")"
 
 FILE="$BUILD_DIR/$NEW_VERSION.zip"
 
-# Update version
-bin/update-boilerplate-dependencies.js
-
-# Install dependencies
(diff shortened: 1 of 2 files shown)


BrunoBernardino (approved):
I haven't tested this locally but it looks like a great improvement!

PR #108, opened 2018-08-07
Ability to extend appRaw at runtime

eliangcs:
Allows `core` to handle the case where `appRawOverride` is an array of `[appRawOverrideHash, appRawExtension]`, where `appRawOverrideHash` is a md5 hash of the app definition when the app definition is too big to send into Lambda, and `appRawExtension` is a partial app definition that will override `appRawOverride` (double override!).

Why? `appRawOverride` is a static app definition so it doesn't contain custom fields that are only determined at runtime. This limits `legacy-scripting-runner` (and also `core`) from identifying file fields in the custom fields. By accepting an extra `appRawExtension` in an array of length 2, the backend sends those custom fields in `appRawExtension` so that `legacy-scripting-runner` is able to simulate file upload with file custom fields in CLI, like what we're doing in this PR: https://github.com/zapier/zapier-platform-legacy-scripting-runner/pull/23.

### DIFF
--- a/integration-test/integration-test.js
+++ b/integration-test/integration-test.js
@@ -1,11 +1,17 @@
 'use strict';
 
-const crypto = require('crypto');
-const should = require('should');
+const fs = require('fs');
+const os = require('os');
 const path = require('path');
-const createLambdaHandler = require('../src/tools/create-lambda-handler');
 
 const AWS = require('aws-sdk');
+const crypto = require('crypto');
+const nock = require('nock');
+const should = require('should');
+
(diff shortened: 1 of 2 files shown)


BrunoBernardino (commented):
I didn't test this locally and the code looks good.

I do have a question, which is why isn't this being merged in the initial override before being sent in the backend?

BrunoBernardino (approved):
Read it through at https://github.com/zapier/zapier/pull/18886 and see this is only for when the merged definition is too big.

PR #109, opened 2018-08-14
Continuation callback support

buildmaster:
In order to support callback functionality the following happens

Zapier passes in a callback url on the raw event object. We take this in the pre-handler and expose a method called `generateCallbackUrl` this allows us to track if the callback url has been used and allows us to perhaps shift the callbackurl generation to RPC in the future.

Once the `generateCallbackUrl` is accessed we can we record the access, and on a post handler we mark the envelope of the return to Zapier as such. This allows Zapier to then follow the callback functionality and pause the task waiting for callback


src/app-middlewares/after/callback-status-catcher.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+let ensureOutputEnvelope = require('../../tools/envelope').ensureOutputEnvelope;
+let STATUSES = require('../../constants').STATUSES;
+const _ = require('lodash');
+/*
+
```
xavdid: were docs going to go in here? if not, we can remove

src/tools/create-callback-wrapper.js
```diff
@@ -0,0 +1,20 @@
...
+//this returns a higher order function that allows users to generate the callback URL
+//generation doesn't actually generate the URL (although we could use RPC if we wanted to)
+//instead it relies on the callbackUrl being passed in by the platform
+//as such it'll always return the same value.
+//calling this inner function will also mark the task as a CALLBACK status on return which will
+//effectively pause the Zap
+
+const createCallbackHigherOrderFunction = input => {
+  let callbackUrl = _.get(input, '_zapier.event.callbackUrl');
+  return function() {
```
xavdid: minor style thing, we mostly use fat arrow syntax, so I'd change to `return () => {`

test/create-app.js
```diff
@@ -516,4 +517,21 @@ describe('create-app', () => {
         .catch(done);
     });
   });
+  describe('calling a callback method', () => {
+    let results;
+    before(() =>
```
xavdid, line 522: this is a personal preference thing, but I don't love the pattern of using a very localized `before`. I think it's more clear to have a function that returns the results. I _think_ mocha on this version is cool with `async // await` but if not, it's fine to leave as is

test/tools/callback-wrapper.js
```diff
@@ -0,0 +1,59 @@
...
+  _zapier: {
+    event: {
+      callbackUrl: CALLBACK_URL
+    }
+  }
+};
+describe('callbackwrapper', () => {
+  let wrapper;
+
+  before(() => {
```
xavdid, line 16: Similar comment in this file - if we can do an `async` function in the `describe` above, let's make a function that calls `createCallbackWrapper` and call that a bunch. way more clear IMO

index.d.ts
```diff
@@ -14,9 +14,9 @@ export const tools: { env: { inject: (filename?: string) => void } };
 export const createAppTester: (
   appRaw: object
 ) => <T extends any, B extends Bundle>(
-  func: (z: ZObject, bundle: B) => Promise<T>,
-  bundle?: Partial<B> // partial so we don't have to make a full bundle in tests
-) => T extends Promise<T> ? T : Promise<T>;
+    func: (z: ZObject, bundle: B) => Promise<T>,
+    bundle?: Partial<B> // partial so we don't have to make a full bundle in tests
+  ) => T extends Promise<T> ? T : Promise<T>;
```
xavdid: super minor, but it looks like the changes here are breaking rank w/ `prettier`? I tried it on my machine and these were reverted. We have a hook that _should_ handle this for you, but maybe it didn't run?

In any case, i'd run this file through `prettier` and re-commit just so we keep that consistent style.

xavdid (commented):
Very cool! I left some minor comments about style but on the whole it looks good. It would be nice to see a test covering the `SUCCESS` handling (if that's important; it's the only thing in code I noticed not being represented in tests). 

I don't think I have enough context on the way this interacts with everything else to 👍, but the code in a CLI vacuum looks good!

test/create-app.js
buildmaster: Thanks for this feedback.

Are you saying you would prefer if each it block called the method and then asserted against the results?

I'm generally against that as it slows down the tests unless you bung a bunch of asserts in the same it block, which then obscures failures and foes you to have to dive into the it block to see what's going on.

I've always followed a describe per scenario, and we're pushing that in cli apps (https://zapier.quip.com/x4SEApNJLlRe)

by which I only mean to add a second opinion, I'm happy to follow the standard laid out before me for this repository, just want to check I'm understanding correctly.

index.d.ts
buildmaster: fixed this up, I must have had some bad prettier settings

test/create-app.js
xavdid: Sorry, I could have been more clear. I was picturing this:

```js
/// instead of this: ///
describe(...
  let results
  before(() => someFunc(...).then(output => {results = output}))
  it('returns a CALLBACK envelope', () => results.status.should.eql('CALLBACK'));
  it('returns method', () => results.status.should.eql('CALLBACK'));
)

/// do this: ///
describe('taking a callback method', () => {
  const genResults = () => app(createTestInput('resources...'))

  it('returns a C...', async () => {
    const results = await genResults()
    results.status.should.eql('CALLBACK'))
  })

  it('returns m...', async () => {
    const results = await genResults()
    results.status.should.eql('CALLBACK'))
  })
})
```

this version makes it a more clear where `results` are coming from. I'm not sure if there are performance considerations (it feels like two pretty similar ways to do the same thing) but if there are, faster is probably better.

src/app-middlewares/after/callback-status-catcher.js
```diff
@@ -0,0 +1,21 @@
...
+by signalling to Zapier that this app/method is returning a callback status the task will be placed
+in a waiting state until the callback is called.
+*/
+let callbackStatusCatcher = output => {
+  const input = output.input || {};
+
+  const callbackUsed = _.get(input, '_zapier.event.callbackUsed');
+  if (callbackUsed) {
+    output = ensureOutputEnvelope(output);
+    output.status = STATUSES.CALLBACK;
```
bcooksey, line 15: Trying to think through how this will interact with [this code](https://github.com/zapier/zapier-platform-core/blob/d9edbf30546b1597a61b6119c4a558195b013de8/src/middleware.js#L49-L55) and having a bit of a hard time wrapping my head around it. Are we safe to assume output will be an envelope at this point?

Otherwise do we need to be concerned about the flag to skip envelope. Also, would that mean that depending on use of `getCallbackUrl()`, any app middlewares that come after this will potentially have two different structures to be concerned with?

Hoping that we can assume it's an envelop and that way it's consistent throughout.

buildmaster: `ensureOutputEnvelope` can be called multiple times as it'll realise it's an envelope shape already and `ensureOutputEnvelope` becomes a no-op.
https://github.com/zapier/zapier-platform-core/blob/d9edbf30546b1597a61b6119c4a558195b013de8/src/middleware.js#L91

says that ensureEnvelope is called before processing middleware, so I think you may be correct that callback-status-catcher should perhaps ensure that options. skipEnvelope isn't set. When is that `true`? 

We already have this case that if `options. skipEnvelope` is `true` your middleware will be called with just the results but if it's false it'll be called with an envelope

test/create-app.js
buildmaster: @xavdid just to complete the circle on this conversation.  `const results = await genResults()` is called twice in your example which means that function will execute two discrete times, in the before method it gets called once (as the setup for the tests) and then the tests get called on the output from calling it once. Hence why I fallback to using the `before` pattern. I'm ok with changing it but it is more expensive.

xavdid: totally. I think in this case, there's a case either way (and I'm fine for you to leave as is). It would be more of an issue if we were worried about keeping `results` fresh for each `it`, since with this setup, the tests could affect each other. Agree that it doesn't matter here though.

src/app-middlewares/after/callback-status-catcher.js
bcooksey: Alrighty, did a bit of debugging to remind myself how this works. It's a bit confusing because the `applyMiddleware` function is used in two very different contexts.

The first context is for app middleware. In this context, we [create the app](https://github.com/zapier/zapier-platform-core/blob/d9edbf30546b1597a61b6119c4a558195b013de8/src/create-app.js#L43) and are expecting that `ensureEnvelope` is going to run and wrap the output in an envelope, which is [the default behavior](https://github.com/zapier/zapier-platform-core/blob/d9edbf30546b1597a61b6119c4a558195b013de8/src/middleware.js#L45).

The second context for `applyMiddleware` is with http middlewares. In that scenario, we [disable ensureEnvelope](https://github.com/zapier/zapier-platform-core/blob/d9edbf30546b1597a61b6119c4a558195b013de8/src/tools/create-request-client.js#L19). It makes sense, we want developers to get ahold of the response object from `z.request`, we don't want it wrapped in an object.

Given that the `skipEnvelope` is an internal-only thing, and seems to exist solely so we can re-use the `applyMiddleware` function, we should be safe to remove `ensureOutputEnvelope` from this new app middleware. It can assume it gets handed an envelope, since all app middlewares have their output wrapped in an envelope. I guess if this ever got bumped to be first in the list of middlewares applied we could have a problem, but, think we should be fine.

buildmaster:
@xavdid @bcooksey I presume I'm ok to merge this, I think everything has been discussed?

bcooksey:
@buildmaster What do you think about https://github.com/zapier/zapier-platform-core/pull/109#discussion_r232817498

bcooksey (approved):
Looking good!

PR #110, opened 2018-09-10
Allow to disable SSL certificate check

eliangcs:
Addresses https://github.com/zapier/zapier/issues/18677, where the "Disable SSL Certificate Checks" option is not honored by `z.request`.

This PR adds a new before middleware `disableSSLCertCheck` to the `z.request` client if a user checks the "Disable SSL Certificate Checks" in their settings (https://zapier.com/app/settings/advanced).


BrunoBernardino (approved):
This looks great!

src/tools/create-app-request-client.js
```diff
@@ -41,6 +42,11 @@ const createAppRequestClient = (input, options) => {
 
   httpBefores.push(addQueryParams);
 
+  const verify = _.get(input, '_zapier.event.verify');
```
BrunoBernardino: I'd only #suggest renaming this to `verifySSL`, `verifySsl`, or `veryifyCertificate`.

PR #111, opened 2018-09-12
Sign dehydrated payloads

eliangcs:
Fixes zapier/zapier#19547. Also fixes the Lambda live test (i.e. `npm run lambda-integration-test` command) along the way.

### DIFF
--- a/src/tools/environment.js
+++ b/src/tools/environment.js
@@ -27,6 +27,14 @@ const cleanEnvironment = () => {
     delete process.env.AWS_SESSION_TOKEN;
     delete process.env.AWS_SECRET_ACCESS_KEY;
   }
+
+  // Lambda may reuse container, which leaves leftovers in process.env. Let's clean
+  // up Zapier specific stuff here.
+  _.each(process.env, (val, key) => {
+    if (key.startsWith('_ZAPIER')) {
+      delete process.env[key];
+    }
+  });
 };
 
 const localFilepath = filename => {
--- a/src/tools/environment.js
+++ b/src/tools/environment.js
@@ -32,9 +32,9 @@ const cleanEnvironment = () => {
 
   // Lambda may reuse container, which leaves leftovers in process.env. Let's clean
   // up Zapier specific stuff here.
-  for (const name of ENV_VARS_TO_CLEAN) {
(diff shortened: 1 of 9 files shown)


src/tools/environment.js
```diff
@@ -27,6 +29,12 @@ const cleanEnvironment = () => {
     delete process.env.AWS_SESSION_TOKEN;
     delete process.env.AWS_SECRET_ACCESS_KEY;
   }
+
+  // Lambda may reuse container, which leaves leftovers in process.env. Let's clean
+  // up Zapier specific stuff here.
+  for (const name of ENV_VARS_TO_CLEAN) {
```
BrunoBernardino: It might be cleaner (and more functional) to:

`ENV_VARS_TO_CLEAN.map((key) => { delete process.env[key]; });`

test/userapp/index.js
```diff
@@ -335,20 +352,20 @@ const MixedInputFields = {
...
-      label: 'mixed input fields',
+      label: 'Mixed Input Fields',
       description:
         'input fields are static, a sync function, and an async promise'
     },
     operation: {
       inputFields: [
         { key: 'key 1' },
         (z, bundle) => Promise.resolve({ key: bundle.key2 }),
-        Promise.resolve({ key: 'key 3' })
```
BrunoBernardino, line 346: This was allowed before and isn't anymore?

BrunoBernardino (approved):
This looks great. I didn't test locally as it seems you got it all going on, but I do have a (non-blocking) question and a #suggestion.

Love the tests!

src/tools/environment.js
eliangcs: I was thinking `for` loop is faster than `map` and `forEach`. But I guess we could use some tooling to transpile all of them to `for` loops if we really want to squeeze performance. Fixed 7edea2f. I used `forEach` instead since we don't need it to return an array as `map` does.

test/userapp/index.js
eliangcs: I'm not sure if we ever allowed this before. But I'm pretty sure we don't allow this now as it won't even pass the schema validation.

PR #112, opened 2018-09-19
Introduce z.dehydrateFile

eliangcs:
Adds `z.dehydrateFile` to CLI. Requires https://github.com/zapier/zapier/pull/19954.

`z.dehydrateFile` has two function signatures:

* `z.dehydrateFile(func, inputData)`: Nothing different from `z.dehydrate(func, inputData)` except that we're annotating the payload with an "it's for a file" tag. So the backend knows not to hydrate it too early in Zap editor.
* `z.dehydrateFile(url, request, meta)`: We'll run `hydrators.zapierDefaultFileHydrator`, which is injected at runtime as part of the app initialization. `zapierDefaultFileHydrator` provides the default implementation for the most common use cases. It downloads the file from the trigger-side app and uploads it to S3 with `z.stashFile`.

### DIFF
--- a/src/tools/create-file-dehydrator.js
+++ b/src/tools/create-file-dehydrator.js
@@ -1,0 +1,80 @@
+'use strict';
+
+const _ = require('lodash');
+
+const { DEFAULT_FILE_HYDRATOR_NAME } = require('../constants');
+const { DehydrateError } = require('../errors');
+const resolveMethodPath = require('./resolve-method-path');
+const wrapHydrate = require('./wrap-hydrate');
+
+const MAX_PAYLOAD_SIZE = 2048; // most urls cannot be larger than 2,083
+
+const createFileDehydrator = input => {
+  const app = _.get(input, '_zapier.app');
+
+  const dehydrateFileFromRequest = (url, request, meta) => {
+    url = url || undefined;
+    request = request || undefined;
+    meta = meta || undefined;
+    if (!url && !request) {
+      throw new DehydrateError(
+        'You must provide either url or request arguments!'
+      );
+    }
+
+    const inputData = {
+      url,
+      request,
+      meta
+    };
+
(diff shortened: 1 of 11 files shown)


test/hydration.js
```diff
@@ -25,26 +31,20 @@ describe('hydration', () => {
...
         dehydrate(funcToMiss, inputData);
-        '1'.should.eql('2'); // shouldn't pass
-      } catch (err) {
-        err.message.should.containEql(
-          'We could not find your function/array/object anywhere on your App definition.'
-        );
-      }
+      }).should.throw(
+        'We could not find your function/array/object anywhere on your App definition.'
+      );
```
bcooksey, line 47: Nice improvement on these tests!

src/tools/create-file-dehydrator.js
```diff
@@ -0,0 +1,58 @@
...
+      );
+    }
+    return wrapHydrate({
+      type: 'file',
+      method: `hydrators.${DEFAULT_FILE_HYDRATOR_NAME}`,
+      bundle: { url, request, meta }
+    });
+  };
+
+  const dehydrateFileFromFunc = (func, inputData) => {
```
bcooksey, line 29: The `fromRequest` and `fromFunc` threw me off a bit on first read cause it made me go "wait, we should be hydrating from a request or a func." I see now that it's communicating "dehydrate, with request as hydration options" or "dehydrate, with func as hydration options", but that doesn't make a very good method name. Unfortunately don't have anything better, so happy to roll with this for now.

bcooksey (approved):
Tested this locally on my machine, both with a function and a simple URL. Confirmed that we no longer hydrate too eagerly during Zap setup, that the default hydrator works, the function version works, and the caching let the editor keep using the same S3 URL.

Big wins!

PR #113, opened 2018-09-20
Better override for perform method on partial definition

BrunoBernardino:
Allowing the complete override of `perform` and `authentication.test` methods when applying a partial definition, to prevent confusion of "what to execute" if there's a
`url` and a `source`, for example, or a function and a `url` in those.

Added tests to confirm the behavior.

This was noticed as part of https://github.com/zapier/zapier/pull/19935

### DIFF
--- a/src/tools/create-lambda-handler.js
+++ b/src/tools/create-lambda-handler.js
@@ -83,7 +83,6 @@ const getAppRawOverride = (rpc, appRawOverride) => {
 
       if (typeof appRawOverride !== 'string') {
         appRawOverride = extendAppRaw(appRawOverride, appRawExtension);
-        console.log(appRawOverride);
         resolve(appRawOverride);
         return;
       }
--- a/src/tools/create-lambda-handler.js
+++ b/src/tools/create-lambda-handler.js
@@ -36,7 +36,7 @@ const extendAppRaw = (base, extension) => {
     'performSubscribe',
     'performUnsubscribe'
   ];
-  const concatArray = (objValue, srcValue, key) => {
+  const concatArrayAndOverrideKeys = (objValue, srcValue, key) => {
     if (Array.isArray(objValue) && Array.isArray(srcValue)) {
       return objValue.concat(srcValue);
     }
@@ -54,7 +54,7 @@ const extendAppRaw = (base, extension) => {
 
     return undefined;
   };
-  return _.mergeWith(base, extension, concatArray);
+  return _.mergeWith(base, extension, concatArrayAndOverrideKeys);
 };
 
 const getAppRawOverride = (rpc, appRawOverride) => {
--- a/src/tools/create-lambda-handler.js
+++ b/src/tools/create-lambda-handler.js
@@ -26,7 +26,48 @@ const extendAppRaw = (base, extension) => {
     }
     return undefined;
   };
-  return _.mergeWith(base, extension, concatArray);
+  const deletePerformIfBeingOverridden = (_base, _extension) => {
+    // Delete "test" from auth and "perform" from _base's triggers, creates, and searches,
+    // if it exists in _extension, in order to not have multiple possibilities (request, function, source)
+    if (
+      _.get(_extension, ['authentication', 'test']) &&
+      _.get(_base, ['authentication', 'test'])
+    ) {
+      delete _base.authentication.test;
+    }
+
+    const topLevelKeys = ['triggers', 'creates', 'searches'];
+    topLevelKeys.forEach(topLevelKey => {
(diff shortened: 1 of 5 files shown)


src/tools/create-lambda-handler.js
```diff
@@ -26,7 +26,48 @@ const extendAppRaw = (base, extension) => {
...
+    }
+
+    const topLevelKeys = ['triggers', 'creates', 'searches'];
+    topLevelKeys.forEach(topLevelKey => {
+      if (!_.get(_extension, [topLevelKey])) {
+        return;
+      }
+
+      Object.keys(_extension[topLevelKey]).forEach(extensionTopLevelKey => {
+        if (
```
BrunoBernardino: I think prettier actually made this more difficult to read, but 🤷‍♂️

eliangcs (commented):
Pulled and tested. The change works perfectly. I have two #suggestions.

1. I wonder if we could make the code much cleaner by moving all the merge logic into the `customizer` of `_.mergeWith`. Something naive would be:

```js
const extendAppRaw = (base, extension) => {
  const customMerge = (objValue, srcValue, key) => {
    if (Array.isArray(objValue) && Array.isArray(srcValue)) {
      return objValue.concat(srcValue);
    }

    if (
      // Do full replacement when it comes to "authentication.test" and "perform"
      (key === 'perform' || key === 'test') &&
      _.isPlainObject(srcValue) &&
      _.isPlainObject(objValue)
    ) {
      return srcValue;
    }

    return undefined;
  };
  return _.mergeWith(base, extension, customMerge);
};
```

This is not very accurate, but it passes your tests. This `customMerge` tells if a value is an `authentication.test` or `(creates|searches|triggers).KEY.operation.perform` by only checking if the one last key is `test` or `perform`. Maybe we want to look closer into `srcValue` and `objValue` to see if they look like a `RequestSchema`.

2. Those diffs that remove `^`s from `package-lock.json` seem to suggest that you're using an older version of `npm`. But I'm not sure. Can you confirm? If it does come from an older version of `npm`, you'll need to upgrade `npm` and rerun `npm install` to get rid of those noises.

BrunoBernardino:
Thank you so much @eliangcs !

1. That makes sense. The only potential problem there is if a trigger/create/search is named `perform` or `test`, where it will lose data on the final definition, and that can be very problematic. I can totally do that simpler version if we acknowledge that and are OK with it since it's probably a very edge-case.

2. I'm on `npm v5.8.0` for `node v8.11.1`. Is that outdated here?

eliangcs:
@BrunoBernardino 

1. Yeah, so besides the one last key, we should also check if there're fields like `source`, `url`, `method`, `params`, etc to make sure we're not replacing the wrong things. We can use `zapier-platform-schema` to do the check:

```js
const RequestSchema = require('zapier-platform-schema/lib/schemas/RequestSchema');
const FunctionSchema = require('zapier-platform-schema/lib/schemas/FunctionSchema');

const isRequestOrFunction = obj => {
  return (
    RequestSchema.validate(obj).valid || FunctionSchema.validate(obj).valid
  );
};

const extendAppRaw = (base, extension) => {
  const customMerge = (objValue, srcValue, key) => {
    if (Array.isArray(objValue) && Array.isArray(srcValue)) {
      return objValue.concat(srcValue);
    }

    if (
      // Do full replacement when it comes to "authentication.test" and "perform"
      (key === 'perform' || key === 'test') &&
      _.isPlainObject(srcValue) &&
      _.isPlainObject(objValue) &&
      isRequestOrFunction(srcValue) &&
      isRequestOrFunction(objValue)
    ) {
      return srcValue;
    }

    return undefined;
  };
  return _.mergeWith(base, extension, customMerge);
};
```

2. I'm using npm 6.1.0. I believe npm 6 started using loose versions (vs exact versions in npm 5) in package-lock.json. See https://github.com/npm/npm/issues/20434.

BrunoBernardino:
Alright @eliangcs please check now!

src/tools/create-lambda-handler.js
```diff
@@ -19,11 +19,39 @@ const environmentTools = require('./environment');
...
 const extendAppRaw = (base, extension) => {
-  const concatArray = (objValue, srcValue) => {
+  const keysToOverride = [
+    'test',
+    'perform',
+    'performList',
+    'performSubscribe',
+    'performUnsubscribe'
+  ];
+  const concatArray = (objValue, srcValue, key) => {
```
eliangcs: We should probably rename `concatArray` to something else as the function doesn't only do array concatenation anymore.

eliangcs (approved):
👍Looks good! Once you rename that `concatArray` function, I think you're good to go.

PR #115, opened 2018-09-24
Fix typing of Bundle.cleanedRequest for hook requests

### LINKED ISSUE #114: Bundle typing is inaccurate for hook bundles
For hook event `Bundle.cleanedRequest` will contain whatever user sends in hook, so it misses ` | any` type

izhyvaiev:
closes https://github.com/zapier/zapier-platform-core/issues/114


xavdid:
@izhyvaiev I'm not sure I follow - as far as I can tell from the [readme](https://github.com/zapier/zapier-platform-cli#bundlecleanedrequest) only the 4 typed keys should be present. Are you seeing different behavior?

izhyvaiev:
`perform` operation for rest hooks has completely different syntax, comparing to all other operations.
Docs for it is not very consistent, but example app - https://github.com/zapier/zapier-platform-example-app-rest-hooks/blob/master/triggers/recipe.js accurately shows how it actually works.

In example app, `perform` is mapped to `getRecipe()`, where at lines 42,43 the is an explanation:

> // bundle.cleanedRequest will include the parsed JSON object (if it's not a
> // test poll) and also a .querystring property with the URL's query string. 

As parsing of valid json can return any of **null, number, string, object or array**, the actual type looks better to be *any*.

All other properties will actually be same as `Bundle`.

jamescarr:
@zapier/security test

xavdid:
thanks! sorry for the delay here.

PR #116, opened 2018-09-25
Add smoke tests

eliangcs:
Smoke tests for core do the following:

1. `npm pack` the core
2. Download and extract an example app
3. In the example app directory, install the core package generated in step 1
4. Make sure both `zapier test` and `zapier build` return 0 (successful exit code)

If both `zapier test` and `zapier build` end successfully, I think we can be fairly confident that the core package contain all the necessary files.

A lot of code was copied from https://github.com/zapier/zapier-platform-cli/pull/361. I don't think there's a good way to reuse the code besides putting all the packages into a monorepo, so let's bear with duplicate code for now.


xavdid (approved):
Sweet. Once they've figured out their process, I think it's worthwhile to adopt [lerna](https://github.com/lerna/lerna) like the front-end team is doing. That way we'll be able to share some of this code. 

Also, maybe worth a comment at the top of the test file that any important changes need to be made in the matching `cli` file.

PR #117, opened 2018-10-05
fix null error handling

xavdid:
Fixes https://github.com/zapier/zapier/issues/18163. 

If a user calls `Promise.reject(null)` in a somewhere, it bubbles to here and causes a further error. This _should_ fix it

I'm not sure how to test this - I would think that it's in `integration-test`, but adding a test that looks a lot like [this](https://github.com/zapier/zapier-platform-core/blob/cc3b6381d098203490deb5714e23c85125a29dd8/test/userapp/index.js#L258-L272) didn't seem to trip like i thought it would. Any ideas would be appreciated!

### DIFF
--- a/src/tools/create-lambda-handler.js
+++ b/src/tools/create-lambda-handler.js
@@ -169,10 +169,10 @@ const createLambdaHandler = appRawOrPath => {
       // the default behavior with callbacks anyway, but don't want
       // to rely on that.
       logger(logMsg, logData).then(() => {
-        if (!constants.IS_TESTING) {
-          err.message +=
-            '\n\nConsole logs:\n' +
-            logBuffer.map(s => `  ${s.message}`).join('');
+        if (!constants.IS_TESTING && err) {
+          err.message += `\n\nConsole logs:\n${logBuffer
+            .map(s => `  ${s.message}`)
+            .join('')}`;
         }
         callbackOnce(err);
       });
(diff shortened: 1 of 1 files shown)


eliangcs (commented):
Pulled and porked around a bit. I think we can't test it because [here](https://github.com/zapier/zapier-platform-core/blob/cc3b6381d098203490deb5714e23c85125a29dd8/integration-test/integration-test.js#L55) we're stopping the error from raising in the integration test runner. I tried testing on production but can't seem to do so. Did you test it on production? If so, what does your example app look like?

xavdid:
Yeah! I'm using CLI app 2174. The one currently pushed has a modified core.

Here's the trigger that shows the error: 

```js
perform: async (z, bundle) => {
    await Promise.reject(null)
    return [{ a: 1 }]
}
```
To see it in action, we want to look at the `console_logs` in GL: 

* [before](https://logs.internal.zapier.com/messages/graylog_1702/d5e83311-cc29-11e8-be39-0a35a90429a4)
* [after](https://logs.internal.zapier.com/messages/graylog_1702/72661480-cc2c-11e8-be39-0a35a90429a4)

So yeah. not sure we can unit test it, but it's fixed!

eliangcs (approved):
Cool! I think we can take a pass on the unit test then!

PR #118, opened 2018-10-10
Fix default file hydrator

eliangcs:
## Background

Default File Hydrator is a function that we inject to the app definition at runtime. It was added in PR #112 to power the newly-introduced `z.dehydrateFile` method. When a dev calls `z.dehydrateFile(url, request, meta)`, they're doing `z.dehydrateFile(hydrators.zapierDefaultHydrator, { url, request, meta })` essentially.

## What This PR Fixes

The `request` and `meta` argument in `z.dehydrateFile(url, request, meta)` are optional. When a dev omits `request` argument, i.e., `z.dehydrateFile(url)`, we should include the user's auth info when we make the request to `url`. On the other hand, if `request` argument is provided, we should **not** include the user's auth info. This is to make the behavior consistent with the Web Builder as described in this [doc](https://zapier.com/developer/documentation/v2/scripting/#dehydrating-files).

There's a doc change in PR https://github.com/zapier/zapier-platform-cli/pull/360 you may want to take a look at: https://github.com/zapier/zapier-platform-cli/pull/360/commits/ff79a0492fe7634f7ce6e18cab182535b8acc22a.

### DIFF
--- a/src/tools/default-file-hydrator.js
+++ b/src/tools/default-file-hydrator.js
@@ -1,11 +1,17 @@
 'use strict';
 
-const defaultFileHydrator = (z, bundle) => {
-  const request = bundle.inputData.request || {};
-  request.url = bundle.inputData.url || request.url;
-  request.raw = true;
+const _ = require('lodash');
 
-  const filePromise = z.request(request);
+const requestInternal = require('./request-client-internal');
+
+const defaultFileHydrator = (z, bundle) => {
+  const requestOptions = bundle.inputData.request || {};
+  const request = _.isEmpty(requestOptions) ? z.request : requestInternal;
+
+  requestOptions.url = bundle.inputData.url || requestOptions.url;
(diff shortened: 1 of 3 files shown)


bcooksey:
I realize that we need the `(url, request, meta)` for WB apps we are going to port over. Do we want to promote that pattern for fresh apps though? It seems we make our lives easier if we say the only proper form is `(func, inputData)`. We could then have the LegacyScriptingRun add it's own definition of `z.dehydrateFile()` that supports the 3 argument version.

Thoughts @xavdid @eliangcs?

xavdid:
Yeah, i like that. we should try to keep them as separate as possible; CLI do things the best way for it and WB can adapt as needed.

eliangcs:
@bcooksey that makes sense. The code is getting complicated here. I'll push another version that movie all of these to legacy-scripting-runner. Closing this PR.

PR #119, opened 2018-10-10
Add Support for App Middlewares

bcooksey:
We designed core to be able to do this, but never added it. Check out [this code comment](https://github.com/zapier/zapier-platform-core/blob/050654422338db4aa165e17b11e0fddc1c253453/src/create-app.js#L20-L21).

This requires https://github.com/zapier/zapier-platform-schema/pull/56 as well.

### DIFF
--- a/src/create-app.js
+++ b/src/create-app.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const applyMiddleware = require('./middleware');
+const ensureArray = require('./tools/ensure-array');
 const schemaTools = require('./tools/schema');
 
 // before middles
@@ -24,10 +25,16 @@ const createApp = appRaw => {
   const frozenCompiledApp = schemaTools.prepareApp(appRaw);
 
   // standard before middlewares
-  const befores = [addAppContext, injectZObject];
+  const befores = [addAppContext, injectZObject].concat(
+    ensureArray(frozenCompiledApp.beforeApp)
+  );
 
   // standard after middlewares
-  const afters = [checkOutput, largeResponseCachePointer, waitForPromises];
+  const afters = [
+    checkOutput,
+    largeResponseCachePointer,
+    waitForPromises
+  ].concat(ensureArray(frozenCompiledApp.afterApp));
 
   const app = createCommandHandler(frozenCompiledApp);
   return applyMiddleware(befores, afters, app);
--- a/src/create-app.js
+++ b/src/create-app.js
@@ -25,16 +25,19 @@ const createApp = appRaw => {
   const frozenCompiledApp = schemaTools.prepareApp(appRaw);
 
   // standard before middlewares
-  const befores = [addAppContext, injectZObject].concat(
-    ensureArray(frozenCompiledApp.beforeApp)
-  );
+  const befores = [
+    addAppContext,
+    injectZObject,
+    ...ensureArray(frozenCompiledApp.beforeApp)
+  ];
 
   // standard after middlewares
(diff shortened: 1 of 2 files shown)


bcooksey:
Curious to get thoughts on this. I believe this was the intended design (setting beforeApp and afterApp on the app itself), but maybe there is a different way I'm forgetting about.

BrunoBernardino (changes requested):
This looks good and while I can't think of a good use case (there are for requests, but we provide before/after request/response), I think it might bring value in the future, and we're not adding much code to support it.

I'd only like to see tests and examples/docs for its usage.

xavdid (commented):
Cool! Much like bruno, I'm not sure what exactly we need this for, but the code looks pretty reasonable. 

If you wanted, you could write it as

```js
const befores = [
  addAppContext, 
  injectZObject,
  ...ensureArray(frozenCompiledApp.beforeApp)
]
```

which feels a little cleaner to me, but it's nbd.

bcooksey:
Cool. Will get this polished up. If you are curious on the usecase that prompted this: For apps that allow custom record types, and then to have dynamic dropdowns for one of those custom record types, we get into a pickle in the Editor. For example, in SugarCRM I could create a record type `Restaurants`, and then on a Create Lead action I could have a field called `FavoriteRestaurant` that I want to be a dynamic dropdown.

When building the app, you run into a problem. You can't create a trigger called `new_restaurants` because you don't know that a user will have that in their account. So you create a generic trigger like `new_record`. But now we have a problem because we need to know what type of record you want us to list, and today the editor does not supply any sort of context when invoking a `prefill`.

Adding this middleware let's you work around this. You can define your `new_record` trigger, but when you specify the field on the Create Lead action, you make the prefill `new_restaurants.id.name`. You then have a `beforeApp` middleware that inspects the method we are supposed to be invoking and can change it to whatever you want. So your middleware can turn `triggers.new_restaurants.perform` into `triggers.new_record.perform` and include `bundle.inputData.resource = 'restaurant'`.

reganstarr:
@bcooksey Looking forward to seeing this ship! Here's how I'm planning on using `beforeApp` for SugarCRM https://github.com/zapier/zapier-platform-app-sugarcrm/pull/5

bcooksey:
@BrunoBernardino Ok, got the tests added.

I'm skipping docs for now, as I've put a big "EXPERIMENTAL" in the schema docs. Want a few Zapiens to try it out and then we can officially open it up

BrunoBernardino (approved):

buildmaster:
@bcooksey I think we need to make a change in zapier/zapier to get this to work

```
web_1           | READ KeyError u'list_dynamic_sla'
web_1           | Traceback (most recent call last):
web_1           |   File "/zapier/api/resources_flow.py", line 1773, in do_get
web_1           |     results = node.api.read_poll(node.action, params=params, meta=new_meta)
web_1           |   File "/zapier/developer_cli/api_base.py", line 1381, in read_poll
web_1           |     if self.read_meta[read_type].get('hook', False):
web_1           | KeyError: u'list_dynamic_sla'
```

from this code: https://github.com/zapier/zapier-platform-app-dynamics/blob/entity_dropdown_support/index.js#L21

BrunoBernardino:
@buildmaster this hasn't been released yet, so it's probably not being used by that app. Or are you using a custom `core` build there?

buildmaster:
I'm using this version of core released to my local docker environment to test it out

PR #121, opened 2018-10-15
PDE-476: Handling omitEmptyParams

BrunoBernardino:
Includes tests and has been tested in a CLI app in production with a modified core.

Requires a schema PR: https://github.com/zapier/zapier-platform-schema/pull/57

Fixes [PDE-476](https://zapierorg.atlassian.net/browse/PDE-476).

### DIFF
--- a/src/http-middlewares/before/add-query-params.js
+++ b/src/http-middlewares/before/add-query-params.js
@@ -2,13 +2,34 @@
 
 const querystring = require('querystring');
 
+// Mutates the object (it'll be deleted later anyway)
+const removeEmptyParams = params =>
+  Object.keys(params).map(param => {
+    if (
+      params[param] === '' ||
+      params[param] === null ||
+      typeof params[param] === 'undefined'
+    ) {
+      delete params[param];
+    }
+  });
+
 // Take params off of req.params and append to url - "?a=1&b=2"".
 // This middleware should run *after* custom middlewares, because
 // custom middlewares might add params.
 const addQueryParams = req => {
   if (Object.keys(req.params || {}).length) {
     const splitter = req.url.indexOf('?') === -1 ? '?' : '&';
-    req.url += splitter + querystring.stringify(req.params);
+
+    if (req.omitEmptyParams === true) {
+      removeEmptyParams(req.params);
+    }
+
+    const stringifiedParams = querystring.stringify(req.params);
+
+    if (stringifiedParams) {
+      req.url += `${splitter}${stringifiedParams}`;
+    }
   }
   delete req.params;
   return req;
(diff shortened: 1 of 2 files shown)


src/http-middlewares/before/add-query-params.js
```diff
@@ -2,13 +2,34 @@
...
+
 // Take params off of req.params and append to url - "?a=1&b=2"".
 // This middleware should run *after* custom middlewares, because
 // custom middlewares might add params.
 const addQueryParams = req => {
   if (Object.keys(req.params || {}).length) {
     const splitter = req.url.indexOf('?') === -1 ? '?' : '&';
-    req.url += splitter + querystring.stringify(req.params);
+
+    if (req.omitEmptyParams === true) {
```
eliangcs: Can this just be `if (req.omitEmptyParams)`?

eliangcs (approved):
The tests look very solid. Thanks for that! Only have a style suggestion (not a blocker). Feel free to merge on your call.

src/http-middlewares/before/add-query-params.js
```diff
@@ -2,13 +2,34 @@
 
 const querystring = require('querystring');
 
+// Mutates the object (it'll be deleted later anyway)
+const removeEmptyParams = params =>
```
xavdid, line 6: I don’t love that this mutates, but given that it’s local and commented, it’s probably fine

src/http-middlewares/before/add-query-params.js
```diff
@@ -2,13 +2,34 @@
...
-    req.url += splitter + querystring.stringify(req.params);
+
+    if (req.omitEmptyParams === true) {
+      removeEmptyParams(req.params);
+    }
+
+    const stringifiedParams = querystring.stringify(req.params);
+
+    if (stringifiedParams) {
+      req.url += `${splitter}${stringifiedParams}`;
```
xavdid, line 31: I think this is fine as is, but I think we’d be better off parsing the url with the stdlib, pulling out the query, merging with new params, and adding the whole thing rather than guessing the splitter and adding it that way. 

Again, it’s _fine_, but using established tools feels a little more above the board.

xavdid (commented):
Looks great! Love the tests. Had some thoughts about url parsing but nothing showstopping. 

One last thing- do we have tests that cover the `splitter` code? It seems like all of the tests (here and previously) call /get, but none with `/get?a=1`. I may be missing it, but I don’t think anyging uses & as the splitter.

src/http-middlewares/before/add-query-params.js
BrunoBernardino: Agreed. I just didn't want to change too much here, honestly.

We only test the `splitter` now by not adding `?` (it was in the case the params existed but where then removed).

src/http-middlewares/before/add-query-params.js
BrunoBernardino: Same here, but since the variable was being mutated/removed before, I went with the less code option here.

PR #123, opened 2018-10-19
Add support for digest auth

eliangcs:
Adds support for HTTP Digest Auth. Example app https://github.com/zapier/zapier-platform-example-app-digest-auth demonstrates the usage with tests.

### DIFF
--- a/src/http-middlewares/before/add-digest-auth-header.js
+++ b/src/http-middlewares/before/add-digest-auth-header.js
@@ -104,8 +104,7 @@ const buildDigestHeader = (username, password, url, method, creds) => {
     creds.qop === 'auth' ||
     creds.qop.split(',').indexOf('auth') >= 0
   ) {
-    cnonce = '575c7f6b0fbcda9c';
-    // cnonce = md5(Date.now().toString()).substr(0, 16);
+    cnonce = md5(Date.now().toString()).substr(0, 16);
     response = md5(`${HA1}:${creds.nonce}:00000001:${cnonce}:auth:${HA2}`);
   } else {
     throw new NotImplementedError(
--- a/src/http-middlewares/before/add-digest-auth-header.js
+++ b/src/http-middlewares/before/add-digest-auth-header.js
@@ -61,21 +61,22 @@ const parseDictHeader = s => {
   const res = {};
 
   const items = parseHttpList(s);
-  for (let item of items) {
-    if (item.indexOf('=') === -1) {
+
+  items.forEach(item => {
+    if (item.includes('=')) {
+      const parts = item.split('=');
+      const name = parts[0];
+      let value = parts.slice(1).join('=');
+
+      if (value.charAt(0) === '"' && value.charAt(value.length - 1) === '"') {
+        value = value.substring(1, value.length - 1);
+      }
+      res[name] = value;
+    } else {
       res[item] = null;
-      continue;
     }
+  });
 
-    const parts = item.split('=');
-    const name = parts[0];
-    let value = parts.slice(1).join('=');
-
-    if (value.charAt(0) === '"' && value.charAt(value.length - 1) === '"') {
-      value = value.substring(1, value.length - 1);
-    }
-    res[name] = value;
-  }
   return res;
 };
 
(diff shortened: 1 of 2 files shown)


src/http-middlewares/before/add-digest-auth-header.js
```diff
@@ -0,0 +1,166 @@
...
+
+const crypto = require('crypto');
+const urllib = require('url');
+
+const fetch = require('node-fetch');
+
+const { NotImplementedError } = require('../../errors');
+
+// This function splits a comma-separated string described by RFC 2068 Section 2.
+// Ported from https://github.com/python/cpython/blob/f081fd83/Lib/urllib/request.py#L1399-L1440
```
xavdid, line 11: Most minor of comments, but I'd pin this to a commit so it still makes sense if they update that file and the function moves around: `https://github.com/python/cpython/blob/f081fd83032be48aefdb1bbcc38ab5deb03785d5/Lib/urllib/request.py#L1399-L1440`

src/http-middlewares/before/add-digest-auth-header.js
```diff
@@ -0,0 +1,166 @@
...
+};
+
+// Parse lists of key, value pairs as described by RFC 2068 Section 2 and convert them
+// into a python dict.
+// Ported from https://github.com/requests/requests/blob/d2962f1d/requests/utils.py#L342-L373
+const parseDictHeader = s => {
+  const res = {};
+
+  const items = parseHttpList(s);
+  for (let item of items) {
```
xavdid: I'd change to `forEach` as not to block the event loop

src/http-middlewares/before/add-digest-auth-header.js
```diff
@@ -0,0 +1,166 @@
...
+    if (item.indexOf('=') === -1) {
```
xavdid: I find `!item.includes('=')` cleaner, but I won't fight about it

xavdid (approved):
a couple of minor things, but otherwise it looks great! thanks a ton!

src/http-middlewares/before/add-digest-auth-header.js
eliangcs: It's already a commit-specific URL. I shortened the commit hash just to make the URL shorter.

src/http-middlewares/before/add-digest-auth-header.js
eliangcs: Fixed in 4f15e89. Thanks!

src/http-middlewares/before/add-digest-auth-header.js
eliangcs: Do you mean `forEach` can be asynchronous? I looked it up, but I can't find any resources suggesting `forEach` being able not to block the event loop. But I fixed it anyway in 4f15e89 just to keep the coding style consistent with elsewhere.

src/http-middlewares/before/add-digest-auth-header.js
xavdid: ah you're very correct!

src/http-middlewares/before/add-digest-auth-header.js
xavdid: I can't find the comment now, but @stevelikesmusic (i think?) had mentioned something about the difference mattering. consistency is good too though!

xavdid:
👍

src/http-middlewares/before/add-digest-auth-header.js
stevelikesmusic: I'm not sure I recall the comment 🤔 

It looks like all of this is synchronous besides the one `fetch`. The only way I know to process these in parallel would be to create `Promises` and then process those with something like `await Promise.all(items.map(asynchronousProcess))`. But that seems like an optimization we don't yet need here? I may be missing some context as well.

Other than what `for..of` (any iterable) and `forEach` (arrays only) can operate on, I think they are pretty much the same in how they execute.

xavdid: ¯\\\_(ツ)_/¯

PR #124, opened 2018-10-23
Fix bug: z.stashFile doesn't pick up filename in Content-Disposition

eliangcs:
Fixes https://github.com/zapier/zapier/issues/14350.

Quoted from the issue:

> Developer uses z.stashFile() but when another actions consumes the file (or downloads from the returned S3 URL) it comes with the default unnamedfile filename and not the filename the URL passed to z.stashFile() is sending in the disposition.

### DIFF
--- a/src/tools/create-file-stasher.js
+++ b/src/tools/create-file-stasher.js
@@ -111,27 +111,36 @@ const createFileStasher = input => {
               let newBufferStringStream = response;
               if (_.isString(response)) {
                 newBufferStringStream = response;
-              } else if (response && response.headers) {
-                if (response.body && typeof response.body.pipe === 'function') {
+              } else if (response) {
+                if (Buffer.isBuffer(response)) {
+                  newBufferStringStream = response;
+                } else if (Buffer.isBuffer(response.dataBuffer)) {
+                  newBufferStringStream = response.dataBuffer;
+                } else if (
+                  response.body &&
+                  typeof response.body.pipe === 'function'
+                ) {
                   newBufferStringStream = response.body;
                 } else {
                   newBufferStringStream = response.content;
                 }
-                knownLength =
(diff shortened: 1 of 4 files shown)


src/tools/create-file-stasher.js
```diff
@@ -140,9 +149,13 @@ const createFileStasher = input => {
                 fileContentType
               );
             };
+
             if (isStreamed) {
               maybeResponse.throwForStatus();
-              return maybeResponse.buffer().then(parseFinalResponse);
+              return maybeResponse.buffer().then(buffer => {
+                maybeResponse.dataBuffer = buffer;
+                return parseFinalResponse(maybeResponse);
```
bcooksey, line 157: As I'm looking at this diff, it seems that before we were passing `parseFinalResponse` a buffer object, but not the complete response object. With the change, now we will be giving a complete response object. Am I understanding that correctly? If so, does that have a side effect for other scenarios where we were streaming files?

Mainly concerned with the fact that we'll now be reading off the headers of the response, where as before we didn't touch them. Seems like before and after, we are still leaving the buffer with the response body untouched, so thinking this will be ok.

bcooksey (approved):

src/tools/create-file-stasher.js
eliangcs: Yes, you understand it correctly. Without this change, we were always ignoring the response headers if the stream mode is on (`z.request({ raw: true })`). I'm not sure about the side effect, though. Do you have an example?

bcooksey: Not specifically. I was thinking it might mess up `.pip()` or `.read()` if somebody was doing that off the response object, but, we've only been passing in the body this whole time, so it shouldn't matter.

PR #126, opened 2018-11-02
Add OAuth1 support

eliangcs:
Adds support for OAuth1. Requires https://github.com/zapier/zapier/pull/20861 and https://github.com/zapier/zapier-platform-schema/pull/59.

OAuth1 has two big differences compared to OAuth2:

* Request Token
* Request Signing

An OAuth1 app will use `getRequestToken` to define how Zapier is going to retrieve the request token in exchange for the access token.

To sign every request following OAuth1 spec, we'll add an `oauth1SignRequest` `beforeRequest` middleware for OAuth1 apps automatically. This middleware takes advantage of the [`auth` property of the request object](https://zapier.github.io/zapier-platform-schema/build/schema.html#requestschema) to generate the necessary signature for OAuth1. `RequestSchema.auth` should be an object that holds the necessary OAuth parameters. Here's an [example](https://github.com/zapier/zapier-platform-example-app-oauth1-tumblr/blob/72a478e65f2126a440d51e609b0bf8ab8610845f/authentication.js#L21-L26).

More example apps (not public yet):

* https://github.com/zapier/zapier-platform-example-app-oauth1-trello
* https://github.com/zapier/zapier-platform-example-app-oauth1-tumblr
* https://github.com/zapier/zapier-platform-example-app-oauth1-twitter


src/execute.js
```diff
@@ -17,7 +19,13 @@ const executeHttpRequest = (input, options) => {
   return input.z
     .request(options)
     .then(throwForStatus)
-    .then(resp => createJSONtool().parse(resp.content));
+    .then(resp => {
+      try {
+        return createJSONtool().parse(resp.content);
+      } catch (err) {
+        return querystring.parse(resp.content);
```
eliangcs, line 25: OAuth1 spec says that the response body returned by `getRequestToken` and `getAccessToken` should be URL encoded. For example, `oauth_token=abcd&oauth_token_secret=1234`. So we need to parse it properly here.

bcooksey: Does this affect existing apps, as we'll now try a new parsing that we didn't before? Also, we won't throw an error for not being able to parse JSON content.

Wondering if we can find a different spot to put this processing to avoid a breaking change.

src/http-middlewares/before/oauth1-sign-request.js
```diff
@@ -0,0 +1,97 @@
...
+      stripQueryFromUrl(req.url),
+      paramsForBaseString,
+      req.auth.oauth_consumer_secret,
+      req.auth.oauth_token_secret
+    );
+
+    // Implements https://tools.ietf.org/html/rfc5849#section-3.5.1
+    req.headers.Authorization = buildAuthorizationHeader(authParams);
+
+    // TODO: Form-encoded body (section 3.5.2) and querystring (3.5.3)?
```
bcooksey, line 92: Are these different spots you can add the auth details?

src/tools/http.js
```diff
@@ -0,0 +1,28 @@
+const _ = require('lodash');
```
bcooksey, line 1: Nice little refactor here to make these more general-use

src/tools/parse-dict-header.js
```diff
@@ -0,0 +1,74 @@
...
+
+  if (part) {
+    res.push(part);
+  }
+
+  return res.map(x => x.trim());
+};
+
+// Parse lists of key, value pairs as described by RFC 2068 Section 2 and convert them
+// into a python dict.
```
bcooksey:
```suggestion
// into an associative array.
```

May be worth swap the "dict" terminology of this file (name of the file and the `parseDictHeader`) into something more common in JS.

src/http-middlewares/before/oauth1-sign-request.js
eliangcs: Yes, you can add them as [body](https://tools.ietf.org/html/rfc5849#section-3.5.2) or [URL query](https://tools.ietf.org/html/rfc5849#section-3.5.3). We probably don't need to implement them as I haven't seen any service providers require these to be in the body or URL query.

src/execute.js
eliangcs: This is a good point! In fa98ad7 I added an `afterResponse` middleware to re-encode form body to JSON. So we're parsing form body only when it's oauth1.

bcooksey: On that re-encoding from form body to JSON, does that only happen for Request Schemas and not for performs that are functions? It'd be super confusing to change up the response type on the dev if they are also going to manipulate the response.

Even if it's only for request schema's, what impact does this have on logging? Will we log JSON when in fact we got form body?

eliangcs: The re-encoding is also going to happen for perform functions, which can be really confusing for devs. I totally agree. Pushed da7d824, where form body parsing only happens for `getRequestAccess` and `getAccessToken` and only when auth type is `oauth1`. Would like to know if I miss anything. Thanks!

bcooksey:
@eliangcs This approach definitely makes more sense than the re-encoding.

Looking at this again brought up a bigger Q: does it make sense to look at the `Content-Type` header and pick the parsing accordingly? I understand the spec says form-encodedIt for this particular case, but it has me wondering if, for Godzilla, we'll want to be a bit smarter with the parsing we do on request templates? Could be a good question to pose to the group.

eliangcs:
@bcooksey some services may not set `Content-Type` to `application/x-www-form-urlencoded` as the spec requires. For instance, Twitter returns a `Content-Type` of `text/html;charset=utf-8`. But I think we can be fuzzier by saying if `json` is not in `Content-Type`, we parse it like a form body. What do you think?

Could you remind me what request templates are being used for in Godzilla? I'm not sure how it'd affect what we're doing here.

bcooksey:
Sorry, I need to be careful with terms. Godzilla is not supporting the `requestTemplate` as is defined in the [CLI schema](https://github.com/zapier/zapier-platform-schema/blob/37fbc3d429c3cf1eff5053325b2d45607ae0f9c7/lib/schemas/AppSchema.js#L47-L50). I meant a request object that goes in a perform.

My rough thought was that we could make [this](https://github.com/zapier/zapier-platform-core/blob/858e59a13664dcb12ead899198e36c00abe55ec0/src/execute.js#L20) a bit smarter to look at the `Content-Type` header and decide what to do, falling back to JSON. I realize you started that approach in an earlier commit where you tried JSON, then tried form-encoded. This is a bit different in that we'd only alter our default behavior of parsing JSON if the Content-Type suggested that we should. I think that's a reasonable behavior compared to a waterfall of parsings we'll put the response body through to try to find one that works.

A benefit to going with an approach of looking at Content-Type is that then all the requests Godzilla apps make would benefit, and it's an easy thing to explain in the docs: "If Content-Type === X, we parse with Y."

You mentioned that Twitter wouldn't work in that case. Maybe that is ok, we'd just have to use code to do the parsing ourselves. If we are worried about backward compatibility, is this a case where legacyScriptingRunner could help?

The guiding principle here is, can we add this in a way where there aren't special cases/processing for OAuth1; at least in CLI and new Godzilla apps. A special case in a converted app is acceptable.

eliangcs:
@bcooksey that makes sense. The [OAuth1 spec](https://tools.ietf.org/html/rfc5849#section-1.2) seems to suggest that the service provider should use `application/x-www-form-urlencoded` in their `Content-Type` header if the response body is form-encoded. We'd probably better follow this standard even though Twitter and Trello don't. Pushed a5bedeb to address this. Also adjusted [oauth1-twitter](https://github.com/zapier/zapier-platform-example-app-oauth1-twitter/blob/master/authentication.js#L44-L48) and [oauth1-trello](https://github.com/zapier/zapier-platform-example-app-oauth1-trello/blob/5e0e35d6ec7824a84c93763b9daca25a5bd68772/authentication.js#L43-L47) example apps accordingly to make it work.

bcooksey (approved):

PR #128, opened 2018-11-12
Update schema version in package-lock.json before publishing

eliangcs:
Fixes a small annoyance where Travis fails on Node 10.x, like this [build](https://travis-ci.org/zapier/zapier-platform-core/builds/430014171). This is because Travis runs `npm ci` (instead of `npm install`), which uses `package-lock.json` (instead of `package.json`), to install dependencies. And the [script](https://github.com/zapier/zapier-platform-core/blob/d9edbf30546b1597a61b6119c4a558195b013de8/bin/bump-dependencies.js) we use to bump schema version only updates `package.json`, leaving `package-lock.json` intact. To fix it, we only need to run `npm install` after we bump the schema version in `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   ],
   "scripts": {
     "preversion": "git pull && npm test",
-    "version": "node bin/bump-dependencies.js && git add package.json",
+    "version": "node bin/bump-dependencies.js && npm install && git add package.json package-lock.json",
     "postversion": "git push && git push --tags",
     "test": "mocha -t 10000 --recursive test",
     "posttest": "npm run lint",


xavdid (approved):
great catch! I think at some point it will be worthwhile to switch to yarn. maybe as part of the work that'll happen for splitting cli/core. Might also be able to re-use some of the lerna stuff the frontend uses. 

Anyway, that's neither here nor there. This is 👍 

I updated the OP to auto close the issue in CLI; hope you don't mind.

eliangcs:
@xavdid I agree. I think eventually we'll move to monorepo layout and manage all the zapier-platform-* repos (cli, core, schema, and the example apps) in one repo. It'll make development a lot easier.

Also, thanks for defaulting to action, but this PR is not related to https://github.com/zapier/zapier-platform-cli/issues/342, so I'm removing it from the OP.

PR #129, opened 2018-11-15
(fix) make sure to censor url encoded values

xavdid:
matches https://github.com/zapier/lambda-wrapper/pull/32 and a while ago. While this wasn't a reported issue here, figure we cover all our bases. See also: https://github.com/zapier/zapier/issues/21418


eliangcs (approved):
Nice catch! I didn't test locally as the test already looks solid. 👍

PR #132, opened 2018-11-22
Re-org legacy scripting source

eliangcs:
Addresses [PDE-538](https://zapierorg.atlassian.net/browse/PDE-538). This allows legacy apps to put scripting source at `.legacy.scriptingSource` in the app definition. The change is backward compatible; the old location `.legacyScriptingSource` will still work.


bcooksey (approved):
Do we need to be backward compatible? I don't think we've had any apps convert that are live, have we?

eliangcs:
@bcooksey the apps generated by `zapier convert` are using `legacyScriptingSource`, so it's probably better to keep it backward compatible.

PR #133, opened 2018-12-18
Allow to skip specific output checks

eliangcs:
Addresses https://github.com/zapier/zapier/pull/22029#pullrequestreview-184255463. Instead of using a hack to bypass the trigger id check, this PR allows devs to use `bundle.skipChecks` to skip certain output checks in core. To skip `triggerHasId` check, for example, the dev would set `bundle.skipChecks` to `['triggerHasId']`.

### DIFF
--- a/src/app-middlewares/after/checks.js
+++ b/src/app-middlewares/after/checks.js
@@ -19,10 +19,15 @@ const checkOutput = output => {
   const runChecks =
     event.method && event.command === 'execute' && !_zapier.skipChecks;
 
+  const bundleSkipChecks = event.bundle.skipChecks || [];
+
   if (runChecks) {
     const rawResults = checks
       .filter(check => {
-        return check.shouldRun(event.method, event.bundle);
+        return (
+          !bundleSkipChecks.includes(check.name) &&
+          check.shouldRun(event.method, event.bundle)
+        );
       })
       .map(check => {
         return check
--- a/test/checks.js
+++ b/test/checks.js
@@ -3,6 +3,7 @@
 require('should');
 
 const checks = require('../src/checks');
+const checkOutput = require('../src/app-middlewares/after/checks');
 
 const isTrigger = require('../src/checks/is-trigger');
 const isSearch = require('../src/checks/is-search');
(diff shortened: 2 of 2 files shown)


bcooksey (commented):
@eliangcs Like this approach a lot more compared to the other. It's a more explicit solution that won't leak anything into app content/output. Appreciate second PR!

How will the `bundle.skipChecks` be set in this design? I'm trying to picture the complete flow of a call from Zapier into Lambda with this.

eliangcs:
@bcooksey this will work with https://github.com/zapier/zapier-platform-legacy-scripting-runner/pull/36. We can set `bundle.skipChecks` any time before the checks.

bcooksey (approved):
K, so will this be visible on the `bundle` that apps themselves take in, like along side `inputData`. Wondering if it makes more sense for this to be on the `event` itself if that is the case. Otherwise there will be this key devs see in the bundle that they don't really care about or need to interact with. Could cause confusion and require us to document something internal.

Not gonna block this, because we could always follow up and change, but, something to think about.

eliangcs:
@bcooksey I thought about making the flag be in the `event`, but legacy-scripting-runner is [limited to have access to `bundle`](https://github.com/zapier/zapier-platform-legacy-scripting-runner/blob/03a44080bfb1db4a9d05085698a41226bf9c2440/index.js#L854). To put the flag in the `event`, we'll have to do it in the backend as opposed to legacy-scripting-runner. Currently, there's no reliable clue for the backend to tell if an app is a converted or a pure CLI app. And I think we should keep it that way - the backend shouldn't be aware of converted apps. So we can do all the legacy WB behaviors in legacy-scripting-runner. Does that make sense?

eliangcs:
Going to merge first because we can revisit and change it.

bcooksey:
@eliangcs Got it. Appreciate the consideration of not letting this leak into too many areas of the codebase. It's a balancing keeping this contained vs not exposing something we don't want apps to have direct control over.

I have a feeling we are going to end up putting something on apps to distinguish converted ones from CLI and newly-built. If we take that step, there may be a way to refactor and not have to rely on bundle.

PR #134, opened 2018-12-19
(fix) don't search-or-create if either is hidden

xavdid:
minor fix for creating a `searchOrCreate` only if both are visible. 

Also some unrelated improvements in that file while i was in there.

### DIFF
--- a/src/tools/schema.js
+++ b/src/tools/schema.js
@@ -7,6 +7,8 @@ const dataTools = require('./data');
 const schemaTools = require('./schema-tools');
 const zapierSchema = require('zapier-platform-schema');
 
+const isVisible = action => !_.get(action, ['display', 'hidden']);
+
 // Take a resource with methods like list/hook and turn it into triggers, etc.
 const convertResourceDos = appRaw => {
   let triggers = {},
@@ -51,13 +53,13 @@ const convertResourceDos = appRaw => {
       creates[create.key] = create;
     }
 
-    if (search && create) {
-      let searchOrCreate = {
+    if (search && create && isVisible(search) && isVisible(create)) {
+      const searchOrCreate = {
         //key: `${resource.key}SearchOrCreate`,
         key: `${search.key}`, // For now this is a Zapier editor limitation (has to match search)
         display: {
           label: `Find or Create ${resource.noun}`,
-          description: (search.display && search.display.description) || ''
+          description: _.get(search, ['display', 'description'], '')
         },
         search: search.key,
         create: create.key
--- a/test/tools/schema.js
+++ b/test/tools/schema.js
@@ -134,6 +134,54 @@ describe('schema', () => {
       });
     });
 
+    it('should not make a searchOrCreate if either is hidden', () => {
+      const dummySearch = {
+        display: {
+          label: 'Find a Foo',
+          description: 'Finds a Foo.'
+        },
+        operation: {
+          perform: () => {
+            return {};
+          }
+        }
+      };
+
+      const dummyCreate = {
+        display: {
+          label: 'Create a Foo',
+          description: 'Creates a Foo.',
+          hidden: true
+        },
+        operation: {
+          perform: () => {
+            return {};
+          }
+        }
+      };
+
+      const appRaw = {
+        resources: {
+          foo: {
+            key: 'foo',
+            noun: 'Foo',
+            search: dummySearch,
+            create: dummyCreate,
+            outputFields: [
+              { key: 'id', type: 'integer' },
+              { key: 'name', type: 'string' }
+            ],
+            sample: {
+              id: 123,
+              name: 'John Doe'
+            }
+          }
+        }
+      };
+      const compiledApp = schema.compileApp(appRaw);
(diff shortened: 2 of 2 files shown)


eliangcs (approved):
I didn't test it with a backend as the test looks solid. Thanks for fixing it!

Hansenq:
Hello! I'm one of the engineers at Affinity, where we've built our own Zapier app using `zapier-platform-core`. I have a note and a request that relate to this PR and will provide more context as to what happened at the bottom.

1. This PR was included in the [minor release](https://semver.org/) of zapier-platform-core [7.6.0](https://github.com/zapier/zapier-platform-cli/blob/master/CHANGELOG.md#core-1) and labeled as a bugfix, but actually causes a breaking change, as was reflected in our Zapier app (and potentially many other companies' apps as well). See below for an explanation.
1. Could we get this PR reverted? Not only would this prevent breaking changes in other companies' Zapier apps when they upgrade to >=7.6.0, we've been using the functionality of displaying a Search or Create to the user without displaying the associated Create. We also want to prevent future users of our app from seeing the Search or Create action in the edit Zap interface (we've technically deprecated them), but we don't want old users' Zaps that do use the action from breaking.

---

Two years ago when we first built our Zapier app, we defined Search, Create, and Get actions on four resources in order to use the automatically-generated Search or Create action. However, we didn't want users to use the normal Create action without first searching for a value (to prevent duplicates), so we set the Create action to `hidden` while keeping the Search action visible. There was no documentation stating that this was prohibited, and we assumed it was expected behavior. Doing so achieved our goal of removing the Create action in favor of the Search or Create actions for our user, and reduced confusion for them on the page where they need to select the action to use.

We updated our `zapier-platform-core` version from 7.2.0 to 7.6.1 last week as part of a minor unrelated fix. According to semantic versioning, there were no breaking changes to the version bump to 7.6.1 (no major version change). The deploy was time sensitive since we needed to address another issue, so we didn't think to test the app before we pushed (this is our bad and we'll make sure to test next time). 

In doing so, all of our customers' zaps that used a Search or Create action stopped working. The zap would error at that step, claiming that it could not find the "<resource-name>Search" step. However, for Zaps that ran the Search or Create action without creating, the action would execute fine--only when the action did not find a matching entry and needed to create one did it fail.

For many hours, we couldn't figure why our upgrade had broken our customers' Zaps, until we decided to check the changelog for zapier-platform-core and found this commit. Once we did so, we un-hid the Create step for our resources and our customers' zaps started working again. However, this doubled the number of actions our customers see when they select our Zap, since now every Search or Create on a resource now had a Create on that resource as well. It also prevents us from deprecating a Search or Create action for new users by hiding it and eventually removing it.

We'd love it if this commit could be reverted to reduce clutter in our Zap and to facilitate deprecation of the Search or Create action without forcing users to reconfigure their zaps. And it would be much appreciated if y'all could be more diligent about marking PRs that introduce breaking changes as a major version bump! If there's a better channel for this feedback, please let me know, and thanks for making such a flexible and powerful product to connect apps to each other!

xavdid:
@Hansenq Hi there! Thanks a bunch for taking the time to write this out. Yours is a perfectly reasonable use case, albeit an unusual one. Before anything else, I wanted to apologize for breaking your app. I'm usually the stickler for sticking to semver and so I'm sad we tripped you up like this.

We talked it over internally and agreed that the automatic behavior that comes with resources (reusing samples, auto-generating triggers, creating `searchOrCreates`, etc) should cover the most common cases, which includes not creating a `searchOrCreate` when part of that is hidden. 

 The good news is that since the entire `resource` structure is just sugar, you've got a couple of options:

* today, you can change your app to depend on core `7.5.0`, hide your creates, and continue happily. Unless there's a feature you're missing from a later release, you can continue using that for a while
* On version `7.6.0` (and later) we don't automatically create the `searchOrCreate`, but you can still create it manually. The structure is [here](https://github.com/zapier/zapier-platform-example-app-search-or-create/blob/fa85268fa5d6f18226090442ff4aa4047f3a6c8c/index.js#L27-L39). The resource keys are `<resource_key>Search` and `<resource_key>Create`. I've just tested and confirmed this works on the latest (and will continue to, since this is the way it's done without resources).

Either of those options should hopefully work. Definitely let me know if that's not the case or there's anything else we can do to help. And again, sorry for the semver slip-up!

Hansenq:
Ahh...gotcha, thanks for the update! It sounds like there are no plans to undo this change, so it seems like we'll probably need to go with option 2 eventually. I'm a bit disinclined to just go with option 1 listed, since another engineer on our team could also misinterpret moving to 7.6.0 as a minor version bump and cause this to happen all over again. 

As long as we use the same keys as our resource search, and move that over to the `searchOrCreates` section, will this be a seamless transition for our users from the resource Search or Create to a manually defined Search or Create? It seems to be the case, but I'd just like to confirm.

We'll implement one of those solutions; hopefully no other companies' apps run into the same issue!

xavdid:
that's correct, the transition will be seamless. once the definition is built, there's no distinction between a search or create generated by resource "compilation" and one created manually. the former is simply sugar on the latter.

PR #137, opened 2019-02-15
[PDE-646] Prune unmatched tokens from request

stevelikesmusic:
## Description
We had added `omitEmptyParams` to core a while back. The problem is that it only looks to see if the param value we send in the request is empty—not if we have a bundle token that has an empty input/replace value:

```js
const req = {
  url: 'https://api.foo.com/things/{{bundle.inputData.thingId}}',
  params: {
    someParam: 'zapier',
    getsIgnored: undefined,
    good: '{{bundle.inputData.good}}'
    bad: '{{bundle.inputData.bad}}'
  }
  input: {
    _zapier: { 
      bundle: { inputData: { good: 'hooray' } }
    }
  }
}
```

In this scenario, we'd get a url that resolves to `'https://api.foo.com/things/%7B%7Bbundle.inputData.thingId%7D%7D?someParam=zapier&good=hooray&bad=%7B%7Bbundle.inputData.bad%7D%7D'` when `omitEmptyParams: true`.

The expected url should be: `'https://api.foo.com/things/?someParam=zapier&good=hooray'`

Another consideration is it looks like we [left](https://github.com/zapier/zapier-platform-core/blob/06d97e3fca867f9007a3d22a171e0b72f0856ec5/src/http-middlewares/before/prepare-request.js#L105) the door [open](https://github.com/zapier/zapier-platform-core/blob/7bed19da1e550b76eda250afe6a8fc999478309e/src/tools/request-client.js#L16) for a `prune` field that would be sent with the request like `replace` or `merge`. Would we want to use this field? I'm not sure where we would set it in the request flow.

Possibly add it in [`requestMerge`](https://github.com/zapier/zapier-platform-core/blob/06d97e3fca867f9007a3d22a171e0b72f0856ec5/src/http-middlewares/before/prepare-request.js#L114) or right after (before we make replacements)?

## Jira Ticket
https://zapierorg.atlassian.net/browse/PDE-646

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -58,6 +58,7 @@ export interface Bundle<InputData = { [x: string]: any }> {
     headers: { [x: string]: string };
     content: { [x: string]: string };
   }>;
+  subscribeData?: { id: string };
   targetUrl?: string;
 }
 
@@ -67,20 +68,24 @@ declare class RefreshAuthError extends Error {}
 
 // copied http stuff from external typings
 export interface HttpRequestOptions {
-  url?: string;
-  method?: HttpMethod;
+  agent?: Agent;
   body?: string | Buffer | ReadableStream | object;
+  compress?: boolean;
+  follow?: number;
+  form?: object;
   headers?: { [name: string]: string };
   json?: object | any[];
+  method?: HttpMethod;
   params?: object;
-  form?: object;
   raw?: boolean;
   redirect?: 'manual' | 'error' | 'follow';
-  follow?: number;
-  compress?: boolean;
-  agent?: Agent;
-  timeout?: number;
+  removeMissingValuesFrom: {
+    params: boolean,
+    body: boolean,
+  };
   size?: number;
+  timeout?: number;
+  url?: string;
 }
 
 interface BaseHttpResponse {
--- a/src/http-middlewares/before/add-query-params.js
+++ b/src/http-middlewares/before/add-query-params.js
@@ -2,27 +2,32 @@
 
 const querystring = require('querystring');
 
+const isCurlies = /{{.*?}}/g;
+const shouldPruneQueryParam = value =>
+  value === '' ||
+  value === null ||
+  value === undefined ||
+  isCurlies.test(value);
+
 // Mutates the object (it'll be deleted later anyway)
-const removeEmptyParams = params =>
-  Object.keys(params).map(param => {
-    if (
-      params[param] === '' ||
-      params[param] === null ||
-      typeof params[param] === 'undefined'
-    ) {
+const pruneMissingQueryParams = params =>
+  Object.keys(params).forEach(param => {
+    if (shouldPruneQueryParam(params[param])) {
       delete params[param];
     }
   });
+
+const hasQueryParams = ({ params = {} }) => Object.keys(params).length;
 
 // Take params off of req.params and append to url - "?a=1&b=2"".
 // This middleware should run *after* custom middlewares, because
 // custom middlewares might add params.
 const addQueryParams = req => {
-  if (Object.keys(req.params || {}).length) {
-    const splitter = req.url.indexOf('?') === -1 ? '?' : '&';
+  if (hasQueryParams(req)) {
+    const splitter = req.url.includes('?') ? '&' : '?';
 
-    if (req.omitEmptyParams) {
-      removeEmptyParams(req.params);
+    if (req.removeMissingValuesFrom.params) {
+      pruneMissingQueryParams(req.params);
     }
 
     const stringifiedParams = querystring.stringify(req.params);
@@ -31,6 +36,7 @@ const addQueryParams = req => {
       req.url += `${splitter}${stringifiedParams}`;
     }
   }
+
   delete req.params;
   return req;
 };
--- a/src/http-middlewares/before/prepare-request.js
+++ b/src/http-middlewares/before/prepare-request.js
@@ -101,8 +101,12 @@ const prepareRequest = function(req) {
 
   req = _.defaults(req, {
     merge: true,
+    // TODO: client will eventually send `removeMissingValuesFrom` rather than `omitEmptyParams`
+    removeMissingValuesFrom: {
+      params: !!req.omitEmptyParams,
+      body: false
+    },
     replace: true, // always replace curlies
-    prune: false, // TODO: do we prune empty {{missing}} vars?,
     _addContext: () => {}
   });
 
--- a/src/tools/cleaner.js
+++ b/src/tools/cleaner.js
@@ -30,27 +30,25 @@ const recurseCleanFuncs = (obj, path) => {
 };
 
 // Recurse a nested object replace all instances of keys->vals in the bank.
-const recurseReplaceBank = (obj, bank) => {
-  bank = bank || {};
+const recurseReplaceBank = (obj, bank = {}) => {
   const replacer = out => {
     if (typeof out !== 'string') {
       return out;
     }
     Object.keys(bank).forEach(key => {
+      // Escape characters (ex. {{foo}} => \\{\\{foo\\}\\} )
       const s = String(key).replace(/[-[\]/{}()\\*+?.^$|]/g, '\\$&');
       const re = new RegExp(s, 'g');
       out = out.replace(re, bank[key]);
     });
+
     return out;
   };
   return recurseReplace(obj, replacer);
 };
 
 // Takes a raw app and bundle and composes a bank of {{key}}->val
-const createBundleBank = (appRaw, event) => {
-  appRaw = appRaw || {};
-  event = event || {};
-
+const createBundleBank = (appRaw, event = {}) => {
   const bundle = event.bundle || {};
   const bank = {
     bundle: _.merge(
--- a/test/create-request-client.js
+++ b/test/create-request-client.js
@@ -373,102 +373,122 @@ describe('request client', () => {
     });
   });
 
-  it('should not omit empty params by default', () => {
-    const request = createAppRequestClient(input);
-    return request({
-      url: 'http://zapier-httpbin.herokuapp.com/get',
-      params: {
-        something: '',
-        cool: 'true'
-      }
-    }).then(responseBefore => {
-      const response = JSON.parse(JSON.stringify(responseBefore));
+  describe('adds query params', () => {
+    it('should not omit empty params by default', () => {
+      const request = createAppRequestClient(input);
+      return request({
+        url: 'http://zapier-httpbin.herokuapp.com/get',
+        params: {
+          something: '',
+          really: '{{bundle.inputData.really}}',
+          cool: 'true'
+        }
+      }).then(responseBefore => {
+        const response = JSON.parse(JSON.stringify(responseBefore));
 
-      response.json.args.something.should.eql('');
-      response.json.args.cool.should.eql('true');
-      response.status.should.eql(200);
+        response.json.args.something.should.eql('');
+        response.json.args.really.should.eql('{{bundle.inputData.really}}');
+        response.json.args.cool.should.eql('true');
+        response.status.should.eql(200);
 
-      const body = JSON.parse(response.content);
-      body.url.should.eql(
-        'http://zapier-httpbin.herokuapp.com/get?something=&cool=true'
-      );
+        const body = JSON.parse(response.content);
+        body.url.should.eql(
+          'http://zapier-httpbin.herokuapp.com/get?something=&really={{bundle.inputData.really}}&cool=true'
+        );
+      });
     });
-  });
 
-  it('should not omit empty params when set as false', () => {
-    const request = createAppRequestClient(input);
-    return request({
-      url: 'http://zapier-httpbin.herokuapp.com/get',
-      params: {
-        something: '',
-        cool: 'true'
-      },
-      omitEmptyParams: false
-    }).then(responseBefore => {
-      const response = JSON.parse(JSON.stringify(responseBefore));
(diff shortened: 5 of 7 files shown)


src/tools/cleaner.js
```diff
@@ -30,27 +30,30 @@ const recurseCleanFuncs = (obj, path) => {
...
       return out;
     }
     Object.keys(bank).forEach(key => {
+      // Escape characters (ex. {{foo}} => \\{\\{foo\\}\\} )
       const s = String(key).replace(/[-[\]/{}()\\*+?.^$|]/g, '\\$&');
       const re = new RegExp(s, 'g');
       out = out.replace(re, bank[key]);
     });
+
+    if (obj.omitEmptyParams) {
```
stevelikesmusic: Do we want to prevent `body` fields from being pruned? Hypothetically, APIs _should_ return the same result if a field is empty vs. omitted?

```
body: { foo: undefined }

// vs

body: {}
```

src/tools/cleaner.js
```diff
@@ -30,27 +30,30 @@ const recurseCleanFuncs = (obj, path) => {
...
+      out = out.replace(unreplacedTokens, '');
+    }
+
     return out;
   };
   return recurseReplace(obj, replacer);
 };
 
 // Takes a raw app and bundle and composes a bank of {{key}}->val
-const createBundleBank = (appRaw, event) => {
```
stevelikesmusic, line 50: Just adding a default where it makes sense

stevelikesmusic:
@xavdid @szchenghuang this still needs a final decision on what to name this feature. Should we:
1. keep `omitEmptyParams`
1. change it to something more specific like `omitEmptyQueryParams`
1. stick with `prune` and use values to specify what to prune/omit (ex. `queryParams`, `body`, `true` (for both), `false` 

We will _never_ prune tokens from `req.url`

[Slack thread](https://zapier.slack.com/archives/C5Z9BP4U9/p1550236692169700) for context

szchenghuang:
This'd be what I guess/interpret them:

- **omitEmptyParams**: looks to see if the param value is empty
- **omitEmptyQueryParams**: looks to see if any query param value is empty

They have different functions, and _omitEmptyParams_ becomes meaningless when _omitEmptyQueryParams_ is present. Therefore I'd prefer _omitEmptyQueryParams_ in place of _omitEmptyParams_.

_prune_ seems promising and covers all conditions. If I understand it correctly, _prune_ is a superset of _omitEmptyQueryParams_, and _omitEmptyQueryParams_ is a superset of _omitEmptyParams_. When a superset is available, its subset is deemed redundant.

– my 2¢.

src/tools/cleaner.js
```diff
@@ -30,27 +30,25 @@ const recurseCleanFuncs = (obj, path) => {
...
-  appRaw = appRaw || {};
-  event = event || {};
-
+const createBundleBank = (appRaw, event = {}) => {
```
eliangcs, line 51: Mutable default arguments can lead to undesired effects when the function changes the argument, even though it doesn't seem to happen to this function. This [doc](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) explains why. It talks about Python, but it also applies to JavaScript.

test/create-request-client.js
```diff
@@ -373,102 +373,122 @@ describe('request client', () => {
...
+        response.json.args.cool.should.eql('true');
+        response.status.should.eql(200);
 
-      const body = JSON.parse(response.content);
-      body.url.should.eql(
-        'http://zapier-httpbin.herokuapp.com/get?something=&cool=true'
-      );
+        const body = JSON.parse(response.content);
+        body.url.should.eql(
+          'http://zapier-httpbin.herokuapp.com/get?something=&really={{bundle.inputData.really}}&cool=true'
```
eliangcs, line 396: It feels like a bug that we send `{{...}}` to a server. I don't see any situation where we want to keep `{{...}}` in a request payload. Maybe it's reasonable to change the default behavior so when the `var` in `{{var}}` is undefined or null, we replace it with an empty string? So here this would become `'http://zapier-httpbin.herokuapp.com/get?something=&really=&cool=true'`. This is going to be a bug fix and a breaking change at the same time.

eliangcs (commented):
Thanks for tackling this! I should've seen we have a `prune` option when reviewing the `omitEmptyParams` PR. 🤦‍♂️ However, my vote here would be:

* We go with the `req.prune` option. Remove `omitEmptyParams` (a breaking change).
* `req.prune` can be `false`, `'params'` (for query params), `'body'` (for body params), or `true` (for query and body params).
* `req.prune` defaults to `false` if it isn't specified.
* `{{missing}}` vars (`undefined` or `null`) are replaced with an empty string when `req.prune` is false.

szchenghuang:
While the name _omitEmptyQueryParams_ does sound more specific than _omitEmptyParams_, it's also a little redundant since request parameters refer to query parameters already. For example, the standard _request_ module in Python just names it _params_ rather than _queryParams_

```
// Python
>>> payload = {'key1': 'value1', 'key2': 'value2'}
>>> r = requests.get('https://httpbin.org/get', params=payload)
>>> print(r.url)
https://httpbin.org/get?key2=value2&key1=value1
```

Under this premise, I'd assume we don't have to limit _omitEmptyParams_ to that **params** variable in the description as a whole.

However, it sounds like _prune_ is a superset of _omitEmptyParams_ and has covered the conditions in question. In that case, its subset would be redundant.

– My 2¢ regarding the parameter naming.

src/tools/cleaner.js
szchenghuang: The change made is just syntax sugar In JS. Converting the linked example in Python to JS,

```
const append_to = (element, to=[]) => {
    to.push(element)
    return to;
}

const my_list = append_to(12)
const my_other_list = append_to(42)

console.log(my_list, my_other_list);
>>  [12], [42]
```

Could indeed be a little surprising coming from other languages.

eliangcs: @szchenghuang thanks! TIL, this is indeed surprising!

stevelikesmusic: And this was a Python TIL for me @eliangcs 😂 Especially the bit about inner functions in closures

xavdid:
I'm with @eliangcs, I like the idea of combining them into a single variable. Rather than have it be `string | boolean` with some magic values, I'd type it as:

```ts
prune {
  queryParams: boolean // I find "params" ambiguous, make it clear that it's querystring
  body: boolean
}
```

This makes the checking code easier to write too. Also, I might lobby for using `omitMissingValuesFrom` instead of `prune`. It's more verbose, but it's more explicit for a user who isn't familiar. The more trips we can save to the docs, the better.

Anyway, make sure you edit `index.d.ts` when you change the user-facing API!

stevelikesmusic:
Ah good points @xavdid! I agree—definitely prefer something more descriptive. So, say for the default, we'll have: 

```js
omitMissingValuesFrom: {
  queryParams: true,
  body: false,
}
```

I didn't think to edit the typescript typings—thanks for pointing that out 👏

xavdid:
yeah, I like it!

stevelikesmusic:
Another question @xavdid @eliangcs: would it make more sense to break this up into two pieces? We could fix the `omitEmptyParams` as it is today, helping devs currently using the tool in the short term. And then submit the breaking change for the next core version bump in a few weeks.

xavdid:
That's fine by me

eliangcs:
> would it make more sense to break this up into two pieces?

@stevelikesmusic sounds good!

> say for the default, we'll have:
> omitMissingValuesFrom: {
>   queryParams: true,
>   body: false,
> }

It's probably better to keep both of them false. Even though it's a breaking change, I think it'd be nice if we keep the default behavior the same. Most of the developers don't use `omitEmptyParams`, so it'd be good not to break their code. That said, we can have a different default in Godzilla.

> I find "params" ambiguous

@xavdid we're using `request.params` for query parameters, so it's more consistent if we also use `params` in `request.prune` (or `request.omitMissingValuesFrom`). I think it's worth trading off the ambiguity for API consistency. Maybe @bcooksey can provide some feedback?

xavdid:
that's fair. I think `parameter` is a generic enough programming term that it's probably worth specifying that these are the ones we send in the `query` string, but it's probably not worth a breaking change for that. Instead, being internally consistent is a good call.

index.d.ts
```diff
@@ -76,15 +76,13 @@ export interface HttpRequestOptions {
   headers?: { [name: string]: string };
   json?: object | any[];
   method?: HttpMethod;
-  merge?: boolean;
```
stevelikesmusic: Overzealously added these internal fields earlier.

src/http-middlewares/before/add-query-params.js
```diff
@@ -7,7 +7,7 @@ const shouldPruneQueryParam = value =>
   value === '' || value == null || isToken.test(value);
 
 // Mutates the object (it'll be deleted later anyway)
-const pruneEmptyQueryParams = params =>
```
stevelikesmusic: Fits better with the name we settled on

src/http-middlewares/before/prepare-request.js
```diff
@@ -101,8 +101,12 @@ const prepareRequest = function(req) {
 
   req = _.defaults(req, {
```
stevelikesmusic, line 102: In the future, we _could_ use `_.defaultsDeep` to be explicit about setting the defaults of a nested object like `removeMissingValuesFrom`.

But, since the defaults are false, if we only get something like:

```js
req: {
  removeMissingValuesFrom: {
    params: true,
  },
}
```

`_.defaults` will skip over `removeMissingValuesFrom` and `body` will be `undefined` which works for our default value of `false`.

All this to say, I initially changed to `_defaultsDeep` 😄, but realized we probably didn't need it.

stevelikesmusic:
@xavdid @eliangcs I think this is what we want for the short term fix. The breaking change can follow up after this goes out. Thanks!

src/http-middlewares/before/add-query-params.js
```diff
@@ -2,27 +2,29 @@
 
 const querystring = require('querystring');
 
+const isToken = /{{.*?}}/g;
+const shouldPruneQueryParam = value =>
+  value === '' || value == null || isToken.test(value);
```
xavdid: ah, the `== null` threw me for a loop here. IMO it's worth making that `=== null || === undefined` for clarity

src/http-middlewares/before/add-query-params.js
```diff
@@ -2,27 +2,29 @@
...
+const hasQueryParams = ({ params = {} }) => Object.keys(params).length;
+
 // Take params off of req.params and append to url - "?a=1&b=2"".
 // This middleware should run *after* custom middlewares, because
 // custom middlewares might add params.
 const addQueryParams = req => {
-  if (Object.keys(req.params || {}).length) {
-    const splitter = req.url.indexOf('?') === -1 ? '?' : '&';
+  if (hasQueryParams(req)) {
+    const splitter = req.url.includes('?') ? '&' : '?';
```
xavdid, line 27: 👍

xavdid (approved):
Sweet! thanks for digging in here.

src/http-middlewares/before/add-query-params.js
stevelikesmusic: That's fine. I was going back and forth on this. I think it's idiomatic for JS, but yeah, we should default to clarity.

test/create-request-client.js
stevelikesmusic: @eliangcs I'll save this for the breaking change. For now, we'll keep the current behavior.

Another consideration is what to replace missing values in the request body. While query params are a little more flexible, `null` vs. `''` in a write operation could be drastically different. Should we always allow a "tokenified" value get passed in the body?

eliangcs: @stevelikesmusic feel free to save this for a separate PR.

Good point on missing values in the request body. I'd expect different behaviors based on the request body format. If it's a JSON body, `null` should still be `null` and should never be serialized to `''`. But if it's a form body, the situation is kind of like the query parameters - we're forced to serialize all the data types to strings, in which case, `null` should become `''`. Does that make sense?

src/http-middlewares/before/add-query-params.js
```diff
@@ -2,27 +2,29 @@
 
 const querystring = require('querystring');
 
+const isToken = /{{.*?}}/g;
```
eliangcs: Maybe `isCurlies` is a better name?

eliangcs (approved):
I have a function name suggestion, but nothing should block this from being merged. Thanks for fixing it!

src/http-middlewares/before/add-query-params.js
stevelikesmusic: I know they are referred to as `tokens` in the frontend and generically throughout Formatic for any things to be replaced by some other value.

I think `isCurlies` does technically describe the regEx. But the regEx is pretty simple, and `isCurlies` doesn't really describe what we're checking for. What about `shouldBeReplaced`? For me, `isCurlies` would convey no more extra context than just using the regEx itself: `/{{.*?}}/g.test(value)`. But I'm not stuck on it either.

eliangcs: Thanks for clarifying! "curlies" is what we call them in the backend. But yeah, it's only a short regex so I don't see it getting confused either way, so it's your call. 👍

stevelikesmusic: Ah ok! If they're "curlies" on the backend, then "curlies" they shall be 😄 I'd rather stick to the same names we're using. Thanks @eliangcs!!

PR #138, opened 2019-02-19
[PDE-723] Resolve subscription related bundle fields

stevelikesmusic:
## Description
In Godzilla, devs have tried setting up hooks using form mode. The problem is that they need to use bundle fields that are not getting resolved via bundleBank in `/cleaner`. 

Now, we'll add `targetUrl` and `subscribeData` to the available bundle fields when resolving tokens from the request.

## Jira
https://zapierorg.atlassian.net/browse/PDE-723

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "version": "node bin/bump-dependencies.js && npm install && git add package.json package-lock.json",
     "postversion": "git push && git push --tags",
     "test": "mocha -t 10000 --recursive test",
+    "test:w": "mocha -t 10000 --recursive test --watch",
     "posttest": "npm run lint",
     "plain-test": "mocha -t 5000 --recursive test",
     "integration-test": "mocha -t 10000 integration-test",
--- a/src/tools/cleaner.js
+++ b/src/tools/cleaner.js
@@ -1,10 +1,18 @@
 'use strict';
 
 const _ = require('lodash');
+const { defaults, pick, pipe } = require('lodash/fp');
 
 const isPlainObj = require('./data').isPlainObj;
 const recurseReplace = require('./data').recurseReplace;
 const flattenPaths = require('./data').flattenPaths;
+
+const DEFAULT_BUNDLE = {
+  authData: {},
+  inputData: {},
+  targetUrl: '',
+  subscribeData: {}
+};
 
 const recurseCleanFuncs = (obj, path) => {
   // mainly turn functions into $func${arity}${arguments}$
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -19,6 +19,7 @@
     "version": "node bin/bump-dependencies.js && npm install && git add package.json package-lock.json",
     "postversion": "git push && git push --tags",
     "test": "mocha -t 10000 --recursive test",
+    "test:w": "mocha -t 10000 --recursive test --watch",
```
stevelikesmusic, line 22: This is a convenience I liked when developing. Can remove it. I found it a pain to have linting pass when I was working on making my tests pass.

.gitignore
```diff
@@ -5,3 +5,4 @@ node_modules
 .DS_Store
 build-boilerplate
 *.tgz
+.vscode
```
stevelikesmusic: Do we want launch templates for vscode in the project? This is helpful for debugging tests. But they are only a quick Google search away as well.

src/tools/cleaner.js
```diff
@@ -46,18 +54,15 @@ const recurseReplaceBank = (obj, bank) => {
...
+const createBundleBank = (appRaw, event = {}) => {
   const bank = {
-    bundle: _.merge(
-      {},
-      { authData: bundle.authData || {} },
-      { inputData: bundle.inputData || {} }
-    ),
+    bundle: finalizeBundle(event.bundle),
     process: {
       env: _.extend({}, process.env || {})
```
stevelikesmusic, line 68: This kind of scares me since any integration can access `process.env`. Are we sure there is no sensitive data in `process.env`. Should we be filtering out any values?

xavdid: discussed in slack, this is fine

.gitignore
xavdid: could be good! that's a good backlog item for the eventual monorepo

xavdid (approved):
awesome!

test/create-request-client.js
```diff
@@ -471,4 +471,47 @@ describe('request client', () => {
...
+      const subscribeInput = createInput({}, event, testLogger);
+      const request = createAppRequestClient(subscribeInput);
+      return request({
+        url: 'http://zapier-httpbin.herokuapp.com/post',
+        method: 'POST',
+        body: {
+          hookUrl: '{{bundle.targetUrl}}'
+        }
+      }).then(response => {
+        const { hookUrl } = JSON.parse(response.request.body);
```
eliangcs: Is it intended? This doesn't check the echoed content from httpbin. Should probably change to:

```suggestion
        const hookUrl = response.json.json.hookUrl;
```

eliangcs (approved):
Looks good! I only have a minor suggestion on the test. Feel free to merge once it's addressed.

test/create-request-client.js
stevelikesmusic: I see 👍. I was copying a similar pattern for validating the body from another test in the suite. Probably a lack of knowledge about what httpbin sends.

Would checking body on the request suffice though? We just want to know that we resolved `{{bundle.targetUrl}}` to the value of `hookUrl`

eliangcs: Since we can be almost 100% certain that httpbin always echoes back the content we sent to it, checking the response from httpbin is probably more reliable than checking `response.request.body`, which is done by us in `core`. Does that make sense?

stevelikesmusic: That does. Thanks for explaining @eliangcs!

PR #139, opened 2019-02-24
[PDE-683] (BREAKING) Resolve curlies to their original data type

stevelikesmusic:
## Description
CLI integrations always use bundle values directly from the bundle we inject to their request functions:

```js
const perform = (z, bundle) => {
  const options = {
    url: 'http://api.example.com/post', 
	method: 'POST',
    body: {
      // We directly access the variable -- names: ['Ron', 'Veronica', 'Brick']
      names: bundle.inputData.names,
    }
  }

  z.request(options);
}
```

But, if we are using curly tokens to be replaced with an actual value, `core` interpolates value in a string. One example is in a perform shorthand RequestSchema.

```js
const perform = {
  url: 'http://api.example.com/post',
  method: 'POST',
  body: {
    // String.prototype.replace will coerce the data type to a string
    // Using the case above -- names: 'Ron,Veronica,Brick' 
    names: '{{bundle.inputData.names}}',
  }
```

This fix keeps the data types of values replaced via curlies. The one exception is when we have an interpolated value as part of a string (ex. `'Bearer {{bundle.authData.access_token}}'`). In this case we:
1. Add to the string as we did before
1. Throw an error if the type is a plain object or array

The tests do a better job of illustrating what we're doing now.

## Jira
https://zapierorg.atlassian.net/browse/PDE-683

### DIFF
--- a/src/tools/cleaner.js
+++ b/src/tools/cleaner.js
@@ -3,9 +3,12 @@
 const _ = require('lodash');
 const { defaults, pick, pipe } = require('lodash/fp');
 
-const isPlainObj = require('./data').isPlainObj;
-const recurseReplace = require('./data').recurseReplace;
-const flattenPaths = require('./data').flattenPaths;
+const {
+  flattenPaths,
+  getObjectType,
+  isPlainObj,
+  recurseReplace
+} = require('./data');
 
 const DEFAULT_BUNDLE = {
   authData: {},
@@ -43,11 +46,35 @@ const recurseReplaceBank = (obj, bank = {}) => {
     if (typeof out !== 'string') {
       return out;
     }
+
     Object.keys(bank).forEach(key => {
       // Escape characters (ex. {{foo}} => \\{\\{foo\\}\\} )
-      const s = String(key).replace(/[-[\]/{}()\\*+?.^$|]/g, '\\$&');
-      const re = new RegExp(s, 'g');
-      out = out.replace(re, bank[key]);
+      const escapedKey = key.replace(/[-[\]/{}()\\*+?.^$|]/g, '\\$&');
+      const matchesKey = new RegExp(escapedKey, 'g');
+
+      if (!matchesKey.test(out)) {
+        return;
+      }
+
+      const matchesCurlies = /({{.*?}})/;
+      const valueParts = out.split(matchesCurlies).filter(Boolean);
+      const replacementValue = bank[key];
+      const isPartOfString = !matchesCurlies.test(out) || valueParts.length > 1;
+      const shouldThrowTypeError =
+        isPartOfString &&
+        (Array.isArray(replacementValue) || _.isPlainObject(replacementValue));
+
+      if (shouldThrowTypeError) {
+        throw new TypeError(
+          `Cannot reliably interpolate objects or arrays into a string. We received an ${getObjectType(
+            replacementValue
+          )}:\n"${replacementValue}"`
+        );
+      }
+
+      out = isPartOfString
+        ? valueParts.join('').replace(matchesKey, replacementValue)
+        : replacementValue;
     });
 
     return out;
--- a/src/tools/data.js
+++ b/src/tools/data.js
@@ -14,6 +14,18 @@ const isPlainObj = o => {
 
 const comparison = (obj, needle) => obj === needle;
 
+const getObjectType = obj => {
+  if (_.isPlainObject(obj)) {
+    return 'Object';
+  }
+
+  if (Array.isArray(obj)) {
+    return 'Array';
+  }
+
+  return _.capitalize(typeof obj);
+};
+
 // Returns a path for the deeply nested haystack where
 // you could find the needle. If the needle is a plain
 // object we try _.isEqual (which could be slow!).
@@ -130,8 +142,7 @@ const recurseExtract = (obj, matcher) => {
 const _IGNORE = {};
 
 // Flatten a nested object.
-const flattenPaths = (data, sep) => {
-  sep = sep || '.';
+const flattenPaths = (data, sep = '.') => {
   const out = {};
   const recurse = (obj, prop) => {
     prop = prop || '';
@@ -179,6 +190,7 @@ module.exports = {
   findMapDeep,
   flattenPaths,
   genId,
+  getObjectType,
   isPlainObj,
   jsonCopy,
   memoizedFindMapDeep,
--- a/test/create-request-client.js
+++ b/test/create-request-client.js
@@ -534,4 +534,117 @@ describe('request client', () => {
       });
     });
   });
+
+  describe('resolves body and header curlies', () => {
+    it('should keep valid data types', () => {
+      const event = {
+        bundle: {
+          inputData: {
+            number: 123,
+            bool: true,
+            float: 123.456,
+            arr: [1, 2, 3]
+          }
+        }
+      };
+      const bodyInput = createInput({}, event, testLogger);
+      const request = createAppRequestClient(bodyInput);
+      return request({
+        url: 'http://zapier-httpbin.herokuapp.com/post',
+        method: 'POST',
+        body: {
+          number: '{{bundle.inputData.number}}',
+          bool: '{{bundle.inputData.bool}}',
+          float: '{{bundle.inputData.float}}',
+          arr: '{{bundle.inputData.arr}}'
+        }
+      }).then(response => {
+        const { json } = response.json;
+
+        json.number.should.eql(123);
+        json.bool.should.eql(true);
+        json.float.should.eql(123.456);
+        json.arr.should.eql([1, 2, 3]);
+      });
+    });
+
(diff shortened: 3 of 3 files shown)


src/tools/cleaner.js
```diff
@@ -43,11 +43,34 @@ const recurseReplaceBank = (obj, bank = {}) => {
     if (typeof out !== 'string') {
       return out;
     }
+
     Object.keys(bank).forEach(key => {
       // Escape characters (ex. {{foo}} => \\{\\{foo\\}\\} )
-      const s = String(key).replace(/[-[\]/{}()\\*+?.^$|]/g, '\\$&');
-      const re = new RegExp(s, 'g');
-      out = out.replace(re, bank[key]);
+      const escapedKey = String(key).replace(/[-[\]/{}()\\*+?.^$|]/g, '\\$&');
```
stevelikesmusic: Changed the names here to be a bit more descriptive. I _think_ we could drop the `String` constructor. Since `key` is derived from `Object.keys`, we should be able to safely assume it is always a string.

src/tools/cleaner.js
```diff
@@ -43,11 +43,34 @@ const recurseReplaceBank = (obj, bank = {}) => {
...
+      const matchesKey = new RegExp(escapedKey, 'g');
+
+      if (!matchesKey.test(out)) {
```
stevelikesmusic: Let's just bail if it's not a match.

src/tools/data.js
```diff
@@ -14,6 +14,18 @@ const isPlainObj = o => {
 
 const comparison = (obj, needle) => obj === needle;
 
+const getObjectType = obj => {
+  if (_.isPlainObject(obj)) {
```
stevelikesmusic, line 18: Any reason we have our own `isPlainObject`? Is it preferred to lodash's?

src/tools/data.js
```diff
@@ -130,8 +142,7 @@ const recurseExtract = (obj, matcher) => {
 const _IGNORE = {};
 
 // Flatten a nested object.
-const flattenPaths = (data, sep) => {
-  sep = sep || '.';
+const flattenPaths = (data, sep = '.') => {
```
stevelikesmusic, line 145: 💅

src/tools/cleaner.js
eliangcs: I think you're right. Let's drop it!

src/tools/cleaner.js
```diff
@@ -43,11 +43,35 @@ const recurseReplaceBank = (obj, bank = {}) => {
...
+        return;
+      }
+
+      const valueParts = out.split(/({{.*?}})/).filter(Boolean);
+      const replacementValue = bank[key];
+      const isPartOfString = valueParts.length > 1;
+
+      if (isPartOfString) {
+        if (_.isArray(replacementValue) || _.isPlainObject(replacementValue)) {
```
eliangcs: I think here we can use native `Array.isArray` instead of lodash's `_.isArray`.

src/tools/cleaner.js
```diff
@@ -43,11 +43,35 @@ const recurseReplaceBank = (obj, bank = {}) => {
...
+              'Cannot interpolate objects or arrays into a string.',
+              `You've sent an ${getObjectType(replacementValue)},`,
+              `which will get coerced to ${replacementValue}`
+            ].join(' ')
+          );
+        }
+      }
+
+      out = isPartOfString
+        ? valueParts.join('').replace(matchesKey, replacementValue)
```
eliangcs, line 76: Looks like `valueParts.join('')` always equals to `out` here. So can we simplify this line to the following?

```suggestion
        ? out.replace(matchesKey, replacementValue)
```

src/tools/data.js
```diff
@@ -14,6 +14,18 @@ const isPlainObj = o => {
 
 const comparison = (obj, needle) => obj === needle;
 
+const getObjectType = obj => {
+  if (_.isPlainObject(obj)) {
+    return 'Object';
+  }
+
+  if (_.isArray(obj)) {
```
eliangcs: Likewise, native `Array.isArray` is preferred.

src/tools/data.js
eliangcs: 🤷‍♂️ but yeah, I think we should stick to lodash's.

src/tools/cleaner.js
```diff
@@ -3,9 +3,9 @@
 const _ = require('lodash');
 const { defaults, pick, pipe } = require('lodash/fp');
 
-const isPlainObj = require('./data').isPlainObj;
-const recurseReplace = require('./data').recurseReplace;
-const flattenPaths = require('./data').flattenPaths;
+const { isPlainObj, getObjectType } = require('./data');
+const { recurseReplace } = require('./data');
+const { flattenPaths } = require('./data');
```
eliangcs: Nice cleanup! Can we make it even better by importing them in one statement?

```js
const { isPlainObj, getObjectType, recurseReplace, flattenPaths } = require('./data');
```

eliangcs (approved):
Tested locally, I don't find any issues. Good work!

There's a failing test you need to fix in `test/logger.js`. Also got some minor suggestions but overall it's a 👍from me. Since it's a breaking change, maybe @bcooksey would like to take a quick look?

src/tools/data.js
eliangcs: 🤷‍♂️ but yeah, I think we should stick to lodash's.

src/tools/cleaner.js
stevelikesmusic: Totally agree. Only used `_.isArray` since we were grabbing `_.isPlainObject` right next to it. No strong feeling about it though

src/tools/cleaner.js
stevelikesmusic: Doh. Nice catch @eliangcs

src/tools/data.js
stevelikesmusic: Yep, same reason as above. Also don't feel strong about it here either.

src/tools/cleaner.js
stevelikesmusic: Haha yikes. I just now noticed that those were coming from the same file. Thanks for that!

src/tools/cleaner.js
```diff
@@ -43,11 +43,35 @@ const recurseReplaceBank = (obj, bank = {}) => {
...
+      const isPartOfString = valueParts.length > 1;
+
+      if (isPartOfString) {
+        if (_.isArray(replacementValue) || _.isPlainObject(replacementValue)) {
+          throw new TypeError(
+            [
+              'Cannot interpolate objects or arrays into a string.',
+              `You've sent an ${getObjectType(replacementValue)},`,
+              `which will get coerced to ${replacementValue}`
+            ].join(' ')
```
xavdid: is there a reason this is an array vs a string that prettier breaks up automatically?

test/create-request-client.js
```diff
@@ -534,4 +534,117 @@ describe('request client', () => {
...
+      const bodyInput = createInput({}, event, testLogger);
+      const request = createAppRequestClient(bodyInput);
+      return request({
+        url: 'http://zapier-httpbin.herokuapp.com/post',
+        method: 'POST',
+        body: {
+          message: 'No arrays, thank you: {{bundle.inputData.badData}}'
+        }
+      }).should.be.rejectedWith(
+        "Cannot interpolate objects or arrays into a string. You've sent an Array, which will get coerced to 1,2,3"
```
xavdid: > coerced to 1,2,3

I find this wording a little confusing, since it's not clear that `1,2,3` is a string. Maybe change the error to be 

`which will get coerced into the string "1,2,3"`

xavdid: also maybe change it to `Cannot reliably interpolate...` to make it clear that we _can_, but it might have weird behavior.

xavdid (approved):
I agree with CH's great review (as per usual). Added a minor gripe on the error message text, but otherwise 👍

src/tools/cleaner.js
stevelikesmusic: Yeah, I just found it easier to read this way than the way that prettier was formatting it. I don't feel strongly about, though. Can totally be just a string.

test/create-request-client.js
stevelikesmusic: Good points

bcooksey: Is `which will get coerced into the string` actually true? We are always going to throw this error, yea? Maybe just a stringified dump of the first 10-20 characters to help them know _which_ array we are getting?

bcooksey (approved):

test/create-request-client.js
stevelikesmusic: Yeah it will always throw. The intent was to say, "Hey there, here's what would have gone through". But maybe that's unnecessary. We could just say, "Can't do it. Here's what you sent".

We are still sending them the data above. But I can see the value in keeping the output simpler.

PR #140, opened 2019-02-24
[PDE-724] (BREAKING) Make removeMissingValuesFrom a first class field from request

stevelikesmusic:
## Description
This probably could have been done in the last PR that addressed this field. But now we're making it a first class field from the client, and yanking `omitEmptyParams`.

## Jira
https://zapierorg.atlassian.net/browse/PDE-724

### DIFF
--- a/src/http-middlewares/before/prepare-request.js
+++ b/src/http-middlewares/before/prepare-request.js
@@ -99,6 +99,7 @@ const finalRequest = req => {
 const prepareRequest = function(req) {
   const input = req.input || {};
 
+  // We will want to use _.defeaultsDeep if one of these nested values ever defaults to true.
   req = _.defaults(req, {
     merge: true,
     removeMissingValuesFrom: {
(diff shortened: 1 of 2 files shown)


src/http-middlewares/before/prepare-request.js
```diff
@@ -101,9 +101,8 @@ const prepareRequest = function(req) {
 
   req = _.defaults(req, {
     merge: true,
-    // TODO: client will eventually send `removeMissingValuesFrom` rather than `omitEmptyParams`
     removeMissingValuesFrom: {
-      params: !!req.omitEmptyParams,
+      params: false,
```
eliangcs, line 106: Maybe we should use `_.defaultsDeep` because we have a nested object?

stevelikesmusic: Had [considered that](https://github.com/zapier/zapier-platform-core/pull/137#discussion_r258030385). I thought we agreed on this approach, but can easily change if you think it's better with `_.defaultsDeep`.

eliangcs: Ah, totally missed it. We can keep using `_.defaults` here, but maybe we can drop a comment here saying something like "if any one of these defaults to true, we'll need to use _.defaultsDeep".

eliangcs (approved):

xavdid (approved):
cool, let's get this in and cut the 8.0 release!

PR #141, opened 2019-03-04
[PDE-765] preserve objects passed to inputData in the bundle bank

stevelikesmusic:
## Description
When we pass values to `inputData` that are line items are dictionaries, they get flattened out. This prevents us from using these two input types in curlies (ex. `{{bundle.inputData.someDict}}`)—the flattened object skips over the top level `someDict` and only exposes its children. 

This fix keeps the behavior we have today and adds the preserved top level object as well.

## Jira
https://zapierorg.atlassian.net/browse/PDE-765

### DIFF
--- a/src/http-middlewares/before/prepare-request.js
+++ b/src/http-middlewares/before/prepare-request.js
@@ -122,8 +122,7 @@ const prepareRequest = function(req) {
   if (req.replace) {
     const bank = cleaner.createBundleBank(
       input._zapier.app,
-      input._zapier.event,
-      { preserve: { 'bundle.inputData': true } }
+      input._zapier.event
     );
     req = cleaner.recurseReplaceBank(req, bank);
   }
--- a/src/http-middlewares/before/prepare-request.js
+++ b/src/http-middlewares/before/prepare-request.js
@@ -122,7 +122,8 @@ const prepareRequest = function(req) {
   if (req.replace) {
     const bank = cleaner.createBundleBank(
       input._zapier.app,
-      input._zapier.event
+      input._zapier.event,
+      { preserve: { 'bundle.inputData': true } }
     );
     req = cleaner.recurseReplaceBank(req, bank);
   }
--- a/src/tools/cleaner.js
+++ b/src/tools/cleaner.js
@@ -88,7 +88,7 @@ const finalizeBundle = pipe(
 );
 
 // Takes a raw app and bundle and composes a bank of {{key}}->val
-const createBundleBank = (appRaw, event = {}, flattenOptions) => {
+const createBundleBank = (appRaw, event = {}) => {
   const bank = {
     bundle: finalizeBundle(event.bundle),
     process: {
@@ -96,7 +96,8 @@ const createBundleBank = (appRaw, event = {}, flattenOptions) => {
     }
   };
 
-  const flattenedBank = flattenPaths(bank, flattenOptions);
+  const options = { preserve: { 'bundle.inputData': true } };
+  const flattenedBank = flattenPaths(bank, options);
   return Object.keys(flattenedBank).reduce((coll, key) => {
     coll[`{{${key}}}`] = flattenedBank[key];
     return coll;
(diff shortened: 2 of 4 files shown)


src/tools/data.js
```diff
@@ -142,14 +142,17 @@ const recurseExtract = (obj, matcher) => {
 const _IGNORE = {};
 
 // Flatten a nested object.
-const flattenPaths = (data, sep = '.') => {
```
stevelikesmusic, line 145: A little refactor to tidy up the function. We never call this function with a different separator, so let's just only use `'.'`. If we do someday flatten objects with a different separator, it will be easy to add since we're now using an options object.

Other changes are using defaults and better native methods like `Object.entries` and `.forEach`. `.map` to me implies we're not mutating an outside objects and will be using the returned array. `.forEach` is a better fit for what we're doing.

src/tools/data.js
```diff
@@ -142,14 +142,17 @@ const recurseExtract = (obj, matcher) => {
...
-      Object.keys(obj).map(key => {
-        const value = obj[key];
-        const newProp = prop ? prop + sep + key : key;
+  const recurse = (obj, prop = '') => {
+    if (_.isPlainObject(obj)) {
+      Object.entries(obj).forEach(([key, value]) => {
+        const newProp = prop ? `${prop}.${key}` : key;
+
+        if (preserve[prop]) {
+          out[newProp] = value;
```
stevelikesmusic, line 153: We _could_ hit eject here and `return` from the loop. But this could be potentially breaking. For now, this change is additive to what we have today.

stevelikesmusic:
http-bin is busted right now. I'll be adding some additional tests once it's up again.

src/tools/data.js
xavdid: if you think you want to break previous behavior, today is a great day for it. last breaking change for at least a little bit. your call.

src/tools/cleaner.js
```diff
@@ -88,15 +88,15 @@ const finalizeBundle = pipe(
 );
 
 // Takes a raw app and bundle and composes a bank of {{key}}->val
-const createBundleBank = (appRaw, event = {}) => {
+const createBundleBank = (appRaw, event = {}, flattenOptions) => {
```
xavdid: As a general style thing, I think we should put any parameters that have defaults at the end of the signature. If I wanted to call this and use the deafult `event`, i'd need to write `createBundleBank(appRaw, undefined, flattenOptions)` which feels cumbersome.

I get that it's used in a pretty specific spot, so maybe it's fine to leave.

xavdid (approved):
looks good! minor style thing, but otherwise let's ship it and get this out the door.

xavdid:
Oh also random, but you can use the "comment" button rather than the "review" button when self-commenting. 



I don't think it actually matters for github process, but it'll make the comment line look less weird. 😁

src/http-middlewares/before/prepare-request.js
```diff
@@ -122,7 +122,8 @@ const prepareRequest = function(req) {
   if (req.replace) {
     const bank = cleaner.createBundleBank(
       input._zapier.app,
-      input._zapier.event
+      input._zapier.event,
```
stevelikesmusic: This is some comment

stevelikesmusic:
It's a weird thing @xavdid. I am clicking on "Comment" and yet it still pops up as reviewed in the timeline 🤷‍♂️

src/tools/data.js
stevelikesmusic: I'm fine with keeping the current behavior. Doesn't seem like there's any reason to break for this.

src/tools/cleaner.js
stevelikesmusic: I generally agree with you @xavdid. I've been in Java land where that pattern was _everywhere_. `flattenOptions` here could have a default as well, but the default is down in `flattenPath` for `options`.

I think `event` has a default in the chance that we don't get it from the integration request. I wouldn't image we'd ever call this function without passing `event`.

We could also just make this a general `options` and have a default as well. I do think that `options` should generally come last.

All that being said, since it _is_ specific, I just moved option down. We should always use these flatten options when creating the bundle bank.

PR #142, opened 2019-03-06
Allow tests to pass with httpbin.org while we figure httpbin

stevelikesmusic:
## Description

The httpbin instance we were using no longer exists on heroku. But, the httpbin instance we're now using company wide has a bug that wraps all string values in arrays.

For now, we'll use the public httpbin.org until we can iron out our internal httpbin.

### DIFF
--- a/integration-test/integration-test.js
+++ b/integration-test/integration-test.js
@@ -244,7 +244,7 @@ const doTest = runner => {
             noun: 'Foo',
             operation: {
               perform: {
-                url: 'https://zapier-httpbin.herokuapp.com/get',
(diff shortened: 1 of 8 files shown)


xavdid (approved):
sweet, i'll merge once tests pass.

I think after this, it's worth putting httpbin in a variable so we can more easily change it in the future, but this is perfect for getting this out the door.

stevelikesmusic:
I had the exact same thought. Tried to doing a regex replace, but there are places where we need to interpolate the string value differently. It'll be a little more involved. Not hard, just tedious.

PR #143, opened 2019-03-06
make removeMissing correctly optional

xavdid:
this should be an optional, not required, param.


stevelikesmusic (approved):
Yep, thanks for catching that @xavdid

PR #144, opened 2019-03-07
Spike for resolving schema version mismatches

stevelikesmusic:
Working out ideas for resolving when an integration's `schema` version doesn't match `core`'s `schema` version. It only focuses on when the `execute` or `request` commands are called.

I'm thinking `validation` should still probably fail since it's used for development.


xavdid:
closing in favor of https://github.com/zapier/zapier/pull/24347

stevelikesmusic:
thanks @xavdid!

PR #145, opened 2019-03-10
Allow Godzilla to require modules in code mode

stevelikesmusic:
## Description
This will allow devs using Godzilla to require modules in code mode. This includes both the Node's core modules like `crypto` and `core`'s own `node_modules`. 

We are preventing usage of certain modules that could pose a security risk or shoot devs in the foot:
- block usage of `fs` or `child_process`
- only allow `lodash` from `core`'s `node_modules`

Would could open up all accessible modules. If a dev is making broad use of `require` in Godzilla though, there's a good chance the integration should be built on CLI.


src/app-middlewares/before/z-object.js
```diff
@@ -20,14 +21,15 @@ const injectZObject = input => {
   const bundle = _.get(input, '_zapier.event.bundle', {});
   const zRoot = {
     console: createLoggerConsole(input),
```
stevelikesmusic, line 22: 🔤 except for `require: createRequire`

src/create-command-handler.js
```diff
@@ -23,7 +24,12 @@ const createCommandHandler = compiledApp => {
     if (!handler) {
       throw new Error(`Unexpected command ${command}`);
     }
-    return handler(compiledApp, input);
+
+    try {
+      return handler(compiledApp, input);
+    } catch (err) {
+      return handleError(err);
+    }
```
stevelikesmusic, line 32: This seemed like the best place to catch errors that resulted from calling a `method` like `operation.perform`. Alternatively, we could catch errors in [`middleware`](https://github.com/zapier/zapier-platform-core/blob/master/src/middleware.js#L90) with a `try/catch` as well. 

It'd be nice to have `method` return a `Promise` and then I think we _might_ be able to catch without the `try/catch`.

Put a box around digging into that part though. For now, I think the best place is either here or `middleware`.

src/errors.js
```diff
@@ -35,4 +36,21 @@ const exceptions = _.reduce(
   {}
 );
 
-module.exports = exceptions;
+const isRequireError = ({ name, message }) =>
+  name === 'ReferenceError' && message === 'require is not defined';
+
```
stevelikesmusic, line 41: I think this function is ok here for now. If we add more to it, then we can break it out to have more generic error handling.

src/tools/schema-tools.js
```diff
@@ -2,10 +2,9 @@
 
 const dataTools = require('./data');
 
-const makeFunction = (source, args) => {
```
stevelikesmusic, line 5: 💅. I started here, but ditched the approach. Left the refactor for better readability.

src/tools/schema-tools.js
```diff
@@ -32,10 +31,7 @@ const findSourceRequireFunctions = appRaw => {
     if (obj && typeof obj === 'object') {
       const numKeys = Object.keys(obj).length;
       if (numKeys === 1 || numKeys === 2) {
-        if (
-          typeof obj.source === 'string' &&
-          obj.source.indexOf('return') !== -1
-        ) {
+        if (typeof obj.source === 'string' && obj.source.includes('return')) {
```
xavdid, line 34: 👍

src/tools/create-require.js
```diff
@@ -0,0 +1,21 @@
+'use-strict';
+
+const { dependencies, devDependencies } = require('../../package.json');
+const { RequireModuleError } = require('../errors');
+
+const createRequire = () => module => {
```
xavdid: While not a reserved word `module` is already defined (eg `module.exports`) and we probably shouldn't reuse it

src/tools/create-require.js
```diff
@@ -0,0 +1,21 @@
...
+  const libs = Object.keys(dependencies).concat(Object.keys(devDependencies));
+  const allowedLibs = ['lodash'];
```
xavdid: since godzilla apps can't depend on things directly, `lodash` can only be required as a happy side effect of `core` already depending on it. not sure there's explicit change needed here, but it was worth mentioning.

src/tools/create-require.js
```diff
@@ -0,0 +1,21 @@
...
+  const disallowedModules = ['fs', 'child_process'];
```
xavdid: given that the point of lambda is to run user code safely, I don't think there's a compelling reason to blacklist any modules right now. worst case, their code doesn't work as expected and they should use CLI instead. especially because godzilla is aimed at less technical users, i'm not sure this gains us anything

src/tools/create-require.js
```diff
@@ -0,0 +1,21 @@
...
+  const disallowedLibs = libs.filter(d => !allowedLibs.includes(d));
+
+  if (disallowedModules.includes(module) || disallowedLibs.includes(module)) {
+    throw new RequireModuleError(
+      `You are attempting to use a disallowed module: "${module}"`
```
xavdid: I would change this error to `..attempting to require a disallowed` since it pinpoints the source of the error at the `require` call

src/tools/create-require.js
```diff
@@ -0,0 +1,21 @@
+'use-strict';
+
+const { dependencies, devDependencies } = require('../../package.json');
+const { RequireModuleError } = require('../errors');
+
+const createRequire = () => module => {
+  const libs = Object.keys(dependencies).concat(Object.keys(devDependencies));
```
xavdid: If an app doesn't list any `devDependencies`, `Object.keys(undefined)` throws an error. Also, I don't believe godzilla apps have any devDeps.

Also, we might want to think about a future where apps don't have to explicitly depend on `core` (if we use lambda layers). I'd import the name from `constants` and list it explicitly in `allowed`

src/errors.js
```diff
@@ -35,4 +36,21 @@ const exceptions = _.reduce(
...
+const handleError = err => {
+  const { RequireModuleError } = exceptions;
+  if (isRequireError(err)) {
+    throw new RequireModuleError(
+      "Node's module system is not in scope. Use z.require() instead."
```
xavdid: I might change the message to be more clear. Remember, we're targeting less technical users. 

* probably use `instead of "require()"` to help them self-service
* I'd change the start to be `For technical reasons, use z.require(...`, since the message about scope doesn't add much

Also if we were real fancy and could easily grab the name of the module they were trying to import, we could do:

`Node's module system is not in scope. Use z.require("lodash") instead of require("lodash").`

xavdid (changes requested):
Sweet! So I really like the direction this went. My biggest suggestion (and the source of my request for changes) is in the `z.require` function.

Given that this is godzilla only (there's no reason a CLI user would ever need this, right?) I think we can safely design for that use case. Namely, it'll be a world where they won't have any dependencies and will only need to `require` stdlib for certain common http-related tasks. Likely `url`, `querystring`, maybe `crypto`. we could also make lodash more explicitly available if we feel the need, but I bet some docs that describe transforming results with `Array.map` will be sufficient.

I don't think there's anything to be gained in making it as restrictive as it is. I would bet that most users will never see those messages anyway

I left some comments in the code that are only relevant if we want to keep the current approach - I honestly think we could ship it as 

```js
const createRequire = m => require(m)
```

I like the idea of blacklisting modules if we needed to, but since it's so easy to use CLI instead (and not have any restrictions) I'm not sure it's worth our time to think about or maintain.

src/tools/create-require.js
stevelikesmusic: Ha! That's an excellent point. 💯

src/tools/create-require.js
stevelikesmusic: Yep, this was all just an idea to toss out. I figured throwing out a restricted version would be easier to show than try explaining.

In going through all this I've been getting a better handle on how apps are loaded and executed in `core`.

src/tools/create-require.js
stevelikesmusic: I think we can make this moot by just going right to `require`

src/tools/create-require.js
stevelikesmusic: I'm all for that. Again, just going right to `require` makes this code moot anyway.

stevelikesmusic:
Thanks for the review @xavdid! I had some questions about what is in scope when `core` runs in lambda, and how CLI apps are executed vs apps using `definition_override`. 

I'm all for just opening up `z.require()`

stevelikesmusic:
Do we even need or want a `createRequire` if all it is is a simple one liner function?

src/tools/create-require.js
xavdid: totally! i've learned a lot too, I really appreciate you going through this. It's some pretty wild stuff!

xavdid:
I'm not sure we do need an explicit function. It's good how it is now and we can expand it there as needed. 

I'm not sure if you've got other questions but I'm happy to field them best I can!

xavdid (approved):

stevelikesmusic:
Thanks @xavdid! I do have a few more questions, but we can chat tomorrow. They don't really impact this merging.

src/tools/create-require.js
xavdid: @stevelikesmusic so I think i've flopped here - it's probably worth blocking anything in the `core` requirements. _maybe_ an exception for lodash since we're unlikely to not depend on it and it's super useful, but that might cause more confusion than help.

stevelikesmusic: yeah I think restricting to just core node is the way to go. If you know lodash, or find yourself reaching or it, you should probably be on the CLI.

xavdid: very fair. I'll grab this next week and ping you on the review. thanks!

PR #146, opened 2019-03-12
fix individual types for removeMissingValuesFrom object

xavdid:
I swear I'll get this. In addition to the object itself being optional, each property on it is optional.


stevelikesmusic:
😆 my miss too

stevelikesmusic (approved):

PR #147, opened 2019-03-28
Add optional parameter to createAppTester to customize storeKey

michellechu77:
- Updated index.d.ts

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -13,7 +13,7 @@ export const tools: { env: { inject: (filename?: string) => void } };
 // see: https://github.com/zapier/zapier-platform-cli/issues/339#issue-336888249
 export const createAppTester: (
   appRaw: object,
-  storeKey: string
+  customStoreKey?: string
 ) => <T extends any, B extends Bundle>(
   func: (z: ZObject, bundle: B) => Promise<T>,
   bundle?: Partial<B> // partial so we don't have to make a full bundle in tests
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,7 +12,8 @@ export const tools: { env: { inject: (filename?: string) => void } };
 
 // see: https://github.com/zapier/zapier-platform-cli/issues/339#issue-336888249
 export const createAppTester: (
-  appRaw: object
+  appRaw: object,
+  storeKey: string
 ) => <T extends any, B extends Bundle>(
   func: (z: ZObject, bundle: B) => Promise<T>,
   bundle?: Partial<B> // partial so we don't have to make a full bundle in tests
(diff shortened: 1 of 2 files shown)


michellechu77:
@xavdid, wasn't too familiar with the index.dt.s file format, but took a guess at it.

index.d.ts
```diff
@@ -12,7 +12,8 @@ export const tools: { env: { inject: (filename?: string) => void } };
 
 // see: https://github.com/zapier/zapier-platform-cli/issues/339#issue-336888249
 export const createAppTester: (
-  appRaw: object
+  appRaw: object,
+  storeKey: string
```
xavdid: very good guess! The only change is that it should be `storeKey?: string` because it's optional.

src/tools/create-app-tester.js
```diff
@@ -36,7 +36,7 @@ const promisifyHandler = handler => {
 };
 
 // A shorthand compatible wrapper for testing.
-const createAppTester = appRaw => {
+const createAppTester = (appRaw, storeKey = null) => {
```
xavdid: this is good, but typically it's good to avoid `null`. It behaves weirdly with default args. 

That said, you don't actually have to set a default here. The verbose version of this function signature is `(appRaw, storeKey=undefined)`, but since js doesn't really enforce arguments at all, it's most clearly written as `(appRaw, storeKey)`. Does that make sense?

xavdid: I see now that you likely copied the `null` from below. Honestly I shouldn't have written it like that in the first place. 😁

src/tools/create-app-tester.js
```diff
@@ -46,14 +46,15 @@ const createAppTester = appRaw => {
     bundle = bundle || {};
 
     const method = resolveMethodPath(appRaw, methodOrFunc);
+    let myStoreKey = storeKey ? `testKey-${storeKey}` : `testKey-${method}-${randomSeed}`
```
xavdid: though it's perfect for quick adhoc use, i'd probably name this variable something else. maybe `customStoreKey`.

That said, looking below, I'd probably move the ternary up here and rename that variable. So the whole thing would be

```js
(appRaw, customStoreKey) => 
// ...
const storeKey = shouldPaginate(appRaw, method)
  ? storeKey
    ? `testKey-${storeKey}`
    : `testKey-${method}-${randomSeed}`
  : undefined;

// ...
  method,
  bundle,
  storeKey
};
```

if you go that route, remember to adjust the `index.d.ts` appropriately.

xavdid (approved):
great job! 

I left a couple of minor style comments. Feel free to address as you see fit and then merge when you're ready.

index.d.ts
michellechu77: 👍

src/tools/create-app-tester.js
michellechu77: good idea, I like your suggested variable name better.

PR #148, opened 2019-04-11
pass correct storekey

xavdid:
the important piece here is that `event.storeKey` exists. I couldn't figure out how to get a good test for this (it's pretty embedded in there) but I might add a smoke test at some point.

### DIFF
--- a/src/tools/create-app-tester.js
+++ b/src/tools/create-app-tester.js
@@ -46,17 +46,17 @@ const createAppTester = (appRaw, customStoreKey) => {
     bundle = bundle || {};
 
     const method = resolveMethodPath(appRaw, methodOrFunc);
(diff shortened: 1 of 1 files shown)


src/tools/create-app-tester.js
```diff
@@ -46,17 +46,17 @@ const createAppTester = (appRaw, customStoreKey) => {
...
+        ? `testKey-${customStoreKey}`
+        : `testKey-${method}-${randomSeed}`
+      : null;
 
     const event = {
       command: 'execute',
       method,
       bundle,
-      customStoreKey,
+      storeKey
```
michellechu77, line 59: We found out that the event is expecting an explicit 'storeKey'.

michellechu77 (approved):
Thanks for remembering to make this change! LGTM

PR #149, opened 2019-04-16
Add meta to the bundle bank for resolving curlies.

stevelikesmusic:
## Description
Meta was left out when adding the ability to resolve curlies. Here we add `meta` to the bundle defaults so it gets added to the bank. Also updated our typedef with the new meta fields.

It should be noted that we are still omitting `rawRequest`, `inputDataRaw`, and `cleanedRequest`, as I believe we decided these were only useful in scripting. Does that sound right @xavdid?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -37,12 +37,10 @@ export interface Bundle<InputData = { [x: string]: any }> {
   inputData: InputData;
   inputDataRaw: { [x: string]: string };
   meta: {
-    frontend: boolean;
-    prefill: boolean;
-    hydrate: boolean;
-    test_poll: boolean;
-    standard_poll: boolean;
-    first_poll: boolean;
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -37,12 +37,10 @@ export interface Bundle<InputData = { [x: string]: any }> {
   inputData: InputData;
   inputDataRaw: { [x: string]: string };
   meta: {
-    frontend: boolean;
-    prefill: boolean;
-    hydrate: boolean;
-    test_poll: boolean;
-    standard_poll: boolean;
-    first_poll: boolean;
+    isFillingDynamicDropdown: boolean;
```
xavdid, line 40: 😬 🙈

xavdid (approved):
good catch!

PR #151, opened 2019-04-19
Censor auth headers

xavdid:
fixes https://github.com/zapier/zapier/issues/25575.

@codebycaleb nailed the root cause - headers were coming into the logging method on strings, so our fancy recursing for sensitive looking keys didn't actually see the keys. This fixes it so we only only log the headers if we know they arrived as objects.

## Testing

1. `zapier init censortest --template=custom-auth`
2. replace the middleware in `index.js` with the following:

```js
const includeApiKey = (request, z, bundle) => {
  if (bundle.authData.apiKey) {
    const basicHash = Buffer.from(`:${bundle.authData.apiKey}`).toString('base64')
    request.headers.Authorization = `Basic ${basicHash}`
  }
  return request
}
```

3. `zapier push`
4. try to create an auth with any random string (`blahasdfhashdf`); it should fail
5. In GL for the request to `https://auth-json-server.zapier.ninja/me`, you'll see the auth header uncensored (`Authorization: basic somethinghere==`)
5. in your app's `package.json`, change the `zapier-platform-core` dependency to `zapier/zapier-platform-core#censor-auth-headers`; `npm i`
8. change the line in `package.json` back to `8.0.1`
8. run `SKIP_NPM_INSTALL=1 zapier push`
4. connect an auth again, use any random string
3. in GL, the request should now have the header properly censored (`Authorization: censored:asdfdfsa:9`)

### DIFF
--- a/src/http-middlewares/after/log-response.js
+++ b/src/http-middlewares/after/log-response.js
@@ -1,31 +1,10 @@
 'use strict';
 
-const prepHeaders = headers => {
-  headers = headers || {};
-
-  if (headers && typeof headers.forEach === 'function') {
-    const _headers = {};
-    headers.forEach((value, name) => {
-      _headers[name] = value;
-    });
-    headers = _headers;
-  }
-
-  return Object.keys(headers).map(k => {
-    const v = headers[k];
-    return `${k}: ${v}`;
-  });
-};
-
 // Prepare a request/reponse to be logged to the backend.
 // Generally respects the "Zapier" request and resp object format.
 const prepareRequestLog = (req, resp) => {
   req = req || {};
   resp = resp || {};
-
-  const requestHeaders = prepHeaders(req.headers);
-
-  const respHeaders = prepHeaders(resp.headers);
 
   let body;
   if (!req.raw) {
@@ -43,11 +22,11 @@ const prepareRequestLog = (req, resp) => {
     request_type: 'devplatform-outbound',
     request_url: req.url,
     request_method: req.method || 'GET',
-    request_headers: requestHeaders.join('\n'),
+    request_headers: req.headers,
     request_data: req.body,
     request_via_client: true,
     response_status_code: resp.status,
-    response_headers: respHeaders.join('\n'),
+    response_headers: resp.headers,
     response_content: body
   };
 
--- a/src/tools/create-logger.js
+++ b/src/tools/create-logger.js
@@ -87,6 +87,7 @@ const makeSensitiveBank = (event, data) => {
         const censored = hashing.snipify(val);
         bank[val] = censored;
         bank[encodeURIComponent(val)] = censored;
+        bank[Buffer.from(val).toString('base64')] = censored;
       }
       return bank;
     },
--- a/src/tools/create-logger.js
+++ b/src/tools/create-logger.js
@@ -11,6 +11,18 @@ const constants = require('../constants');
 
 const truncate = str => dataTools.simpleTruncate(str, 3500, ' [...]');
 
+const formatHeaders = (headers = {}) => {
+  if (_.isString(headers)) {
+    // we had a bug where headers coming in as strings weren't getting censored. If something calls this with stringified headers, we'll bow out. Pass the raw object instead.
+    return 'WARNING: possibly uncesored headers';
+  }
+  return Object.entries(headers)
+    .map(([header, value]) => {
+      return `${header}: ${value}`;
+    })
+    .join('\n');
+};
+
 // format HTTP request details into string suitable for printing to stdout
 const httpDetailsLogMessage = data => {
   if (data.log_type !== 'http') {
@@ -115,6 +127,9 @@ const sendLog = (options, event, message, data) => {
       safeData[key] = unsafeData[key];
     }
   });
+
+  safeData.request_headers = formatHeaders(safeData.request_headers);
+  safeData.response_headers = formatHeaders(safeData.response_headers);
 
   const body = {
     message: safeMessage,
--- a/src/tools/create-logger.js
+++ b/src/tools/create-logger.js
@@ -14,7 +14,7 @@ const truncate = str => dataTools.simpleTruncate(str, 3500, ' [...]');
 const formatHeaders = (headers = {}) => {
   if (_.isString(headers)) {
     // we had a bug where headers coming in as strings weren't getting censored. If something calls this with stringified headers, we'll bow out. Pass the raw object instead.
-    return 'WARNING: possibly uncesored headers';
+    return 'ERR - refusing to log possibly uncesored headers';
   }
   return Object.entries(headers)
     .map(([header, value]) => {
--- a/src/tools/create-logger.js
+++ b/src/tools/create-logger.js
@@ -8,19 +8,31 @@ const dataTools = require('./data');
 const hashing = require('./hashing');
 const ZapierPromise = require('./promise');
 const constants = require('../constants');
+const { unheader } = require('./http');
 
 const truncate = str => dataTools.simpleTruncate(str, 3500, ' [...]');
 
 const formatHeaders = (headers = {}) => {
+  if (_.isEmpty(headers)) {
+    return undefined;
+  }
   if (_.isString(headers)) {
     // we had a bug where headers coming in as strings weren't getting censored. If something calls this with stringified headers, we'll bow out. Pass the raw object instead.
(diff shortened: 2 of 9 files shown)


src/tools/create-logger.js
```diff
@@ -87,6 +99,7 @@ const makeSensitiveBank = (event, data) => {
         const censored = hashing.snipify(val);
         bank[val] = censored;
         bank[encodeURIComponent(val)] = censored;
+        bank[Buffer.from(val).toString('base64')] = censored;
```
xavdid, line 114: we don't actually need this anymore (this was something I was trying as a solution) but it'll catch if other declared secrets have been base-64 encoded and put in a weird key somewhere (like `?mything=secretkey123`, but the secret is b64 encoded). doesn't hurt to keep ¯\\\_(ツ)_/¯

src/tools/create-logger.js
```diff
@@ -11,6 +11,18 @@ const constants = require('../constants');
 
 const truncate = str => dataTools.simpleTruncate(str, 3500, ' [...]');
 
+const formatHeaders = (headers = {}) => {
+  if (_.isString(headers)) {
+    // we had a bug where headers coming in as strings weren't getting censored. If something calls this with stringified headers, we'll bow out. Pass the raw object instead.
+    return 'ERR - refusing to log possibly uncesored headers';
```
codebycaleb: tiny typo here (and in the tests) - `uncesored` should be `uncensored`

src/tools/create-logger.js
```diff
@@ -11,6 +11,18 @@ const constants = require('../constants');
...
+  }
+  return Object.entries(headers)
```
codebycaleb: Is there an advantage to using `Object.entries(headers).map` over `_.map(headers)`? FWIW I think the equivalent to this would be

```
_.map(headers, (value, header) => `${header}: ${value}`)
```

src/tools/create-logger.js
codebycaleb: 👏 I like that a lot!

test/logger.js
```diff
@@ -76,6 +76,58 @@ describe('logger', () => {
...
+    const logger = createlogger({ bundle }, options);
+
+    return logger('123 from url google.com', bundle.headers).then(response => {
+      response.status.should.eql(200);
+      const j = response.content.json;
+      j.data.request_headers.should.eql(
+        'authorization: basic :censored:10:d98440830f:'
+      );
+      j.data.response_headers.should.eql(
+        'Authorization: :censored:30:f914b1b0d1:'
```
codebycaleb, line 106: This is cool! I really like the way that this worked out where the request header is more specific (includes the `basic` part) because it included the auth value while the response header censored the whole thing because it was a sensitive key. 👍 

I was confused as to why the length for the request header was 10 at first (because `dmVyeXNlY3JldA==` is 16 characters). It turns out this must be a side-effect from https://github.com/zapier/zapier-platform-core/pull/151/files#diff-0936f4d83f660e8d3b1ba19e2225e4c6R102 - we're replacing `dmVyeXNlY3JldA==` with the censored version of `verysecret`. I don't think that's inherently bad (we've been doing the same thing for URI-encoded values), just a side-effect to keep in mind. The real key is that the value should be censored, and that's being achieved. 👍

codebycaleb (approved):
I think this is awesome work! It's definitely a much more concise approach than I would have ended up with 😂 . Thanks for taking this on!

codebycaleb:
I just ran through the testing steps!

Before changes: https://cdn.zapier.com/storage/files/9ef40c0013057098d3e6ba359ea00cfa.txt
After changes: https://cdn.zapier.com/storage/files/f9f68acecca97dbee154cf2ab9624f94.txt

Good news: the request header was censored after changes!
Bad news: the response headers got logged as `-headers: [object Object]` :( I think I've spotted what's going on - I'll drop a comment on the file.

src/http-middlewares/after/log-response.js
```diff
@@ -43,11 +22,11 @@ const prepareRequestLog = (req, resp) => {
     request_type: 'devplatform-outbound',
     request_url: req.url,
     request_method: req.method || 'GET',
-    request_headers: requestHeaders.join('\n'),
+    request_headers: req.headers,
     request_data: req.body,
     request_via_client: true,
     response_status_code: resp.status,
-    response_headers: respHeaders.join('\n'),
+    response_headers: resp.headers,
```
codebycaleb, line 29: So, it turns out the weird check for `headers.forEach` in `prepHeaders` was because `req.headers` gets stored as JSON, but `resp.headers` gets stored as a custom `Headers` object (which has a `forEach` proto added):



That being said, there is a `toJSON` method for the `Headers` object:



So I think the following should work:

```suggestion
    response_headers: resp.headers.toJSON(),
```

EDIT: I followed up with making the above change in my `node_modules/zapier-platform-core/` directory for my test app and got the following successful result:

https://cdn.zapier.com/storage/files/ccdc2dc5523a7839afdd8ed755ea0ab7.txt

src/tools/create-logger.js
xavdid: I like lodash a lot but It's also nice to use standard language features when available. ¯\\\_(ツ)_/¯

src/http-middlewares/after/log-response.js
xavdid: ok, so rather than need the folks calling `logger` needing to know to transform their objects, I made logger be more flexible with what it handled. great catch though!

src/tools/create-logger.js
```diff
@@ -37,15 +61,15 @@ const httpDetailsLogMessage = data => {
 ${trimmedData.request_method || 'GET'} ${
     trimmedData.request_url
   }${trimmedData.request_params || ''}
-${trimmedData.request_headers || ''}
+${formatHeaders(trimmedData.request_headers) || ''}
 
-${trimmedData.request_data || ''}
+${maybeStringify(trimmedData.request_data) || ''}
```
xavdid, line 66: these were dropping `[object: Object]` in local logs, so I spruced

src/tools/data.js
```diff
@@ -169,6 +169,9 @@ const flattenPaths = (data, { preserve = {} } = {}) => {
 
 // A simpler, and memory-friendlier version of _.truncate()
 const simpleTruncate = (string, length, suffix) => {
+  if (string === undefined) {
```
xavdid, line 172: this was turning `undefined` into `''`, which the header formatter was yelling about.

test/logger.js
xavdid: awesome look! yeah, you're correct.

xavdid:
ok @codebycaleb this grew a little bit but I've made some good progress on organizing the logging code to be more robust and centralized. If you give it another quick look next week, we'll be in good shape.

test/misc-tools.js
```diff
@@ -58,7 +58,7 @@ describe('Tools', () => {
   it('should truncate many things!', () => {
     const tests = [
       { value: null, length: 5, suffix: undefined, expected: '' },
-      { value: undefined, length: 5, suffix: '...', expected: '' },
+      { value: undefined, length: 5, suffix: '...', expected: undefined },
```
bcooksey, line 61: Why change undefined to return as undefined instead empty string?

xavdid: see [here](https://github.com/zapier/zapier-platform-core/pull/151#discussion_r277093772) - I changed that method.

Because we're doing 

```js
data.response_headers = someFunc(data.response_headers)
```

`data` goes from `{a: 1}` to `{a: 1, response_headers: undefined}`. So, I changed that behavior. avoiding `truncate` turn `undefined` into `''` (which felt unintuitive).

The other option would be only call `someFunc` when `data.response_headers` already exists to avoid adding new properties. This felt cleaner.

bcooksey: Got it. I wasn't making the connection to it adding a prop in a caller. Makes sense!

codebycaleb:
Re-ran through the test:

`8.0.1` result: https://cdn.zapier.com/storage/files/72031204b3fe0af81f2df28d7d27adda.txt
New branch results: https://cdn.zapier.com/storage/files/ef3e4b849d14da76b1e3466c597e02e0.txt

🎉 Looks good there! I'll take a look at the code too to see if I find anything!

codebycaleb (approved):
I think all of these changes look clean and great! Thanks again for the work on this!

PR #153, opened 2019-05-27
Prevent raw curies from being sent in a request. Allow removing empty…

stevelikesmusic:
## Ticket
https://zapierorg.atlassian.net/browse/PDE-889 from Issue
 
From issue: https://github.com/zapier/zapier/issues/25928

This PR addresses two issues:
1. Prevent unresolved curlies from being sent in query params or the request body
1. Check if we should prune missing values from the body

A couple considerations, we don't remove unresolved curlies from url path. This might allow something like `https://api.com/resouce/{{bundle.inputData.id}}`

Some [context](https://zapier.slack.com/archives/C5Z9BP4U9/p1558623857027800) from slack

### DIFF
--- a/src/http-middlewares/before/add-query-params.js
+++ b/src/http-middlewares/before/add-query-params.js
@@ -1,21 +1,7 @@
 'use strict';
 
 const querystring = require('querystring');
-
-const isCurlies = /{{.*?}}/g;
-const shouldPruneQueryParam = value =>
-  value === '' ||
-  value === null ||
-  value === undefined ||
-  isCurlies.test(value);
-
-// Mutates the object (it'll be deleted later anyway)
-const pruneMissingQueryParams = params =>
-  Object.keys(params).forEach(param => {
-    if (shouldPruneQueryParam(params[param])) {
-      delete params[param];
-    }
-  });
+const { normalizeEmptyParamField } = require('../../tools/cleaner');
 
 const hasQueryParams = ({ params = {} }) => Object.keys(params).length;
 
@@ -26,9 +12,7 @@ const addQueryParams = req => {
   if (hasQueryParams(req)) {
     const splitter = req.url.includes('?') ? '&' : '?';
 
-    if (req.removeMissingValuesFrom.params) {
-      pruneMissingQueryParams(req.params);
-    }
+    Object.entries(req.params).forEach(normalizeEmptyParamField(req));
 
     const stringifiedParams = querystring.stringify(req.params);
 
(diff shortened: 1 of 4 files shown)


src/http-middlewares/before/add-query-params.js
```diff
@@ -26,9 +12,7 @@ const addQueryParams = req => {
   if (hasQueryParams(req)) {
     const splitter = req.url.includes('?') ? '&' : '?';
 
-    if (req.removeMissingValuesFrom.params) {
-      pruneMissingQueryParams(req.params);
-    }
+    Object.entries(req.params).forEach(normalizeEmptyParamField(req));
```
stevelikesmusic: Toyed with using `reduce` to return a new object that could get piped to later operations. In the end, we're mutating request many other places, so stuck with that pattern.

src/tools/cleaner.js
```diff
@@ -110,9 +110,39 @@ const createBundleBank = (appRaw, event = {}) => {
 
 const maskOutput = output => _.pick(output, 'results', 'status');
 
+const normalizeEmptyRequestField = (shouldCleanup, field) => req => {
```
stevelikesmusic: Kept this as a callback to an array iterator function like `forEach`. We _could_ pull that part in here and have a simple call to normalize that loops for the supplied field from request.

xavdid: I think I like the idea of centralizing the call to `Object.entries` so we don't need to keep re-typing that. Doesn't look like there's any place where the calling function _needs_ to change that callback directly.

xavdid (approved):
So the code itself looks good, but i'm a little worried about changing default behavior (even if it's bad). I get that we need to abstract this some for godzilla use, but I think we should be careful. ¯\\\_(ツ)_/¯

Also, you sure do like your function factories 😀

stevelikesmusic:
I go back and forth between factories and partial application 😆 Definitely a fan of some type of curried pattern.

I agree about being careful. What would you suggest? This is a CLI problem as well. Godzilla is just a CLI perform using the request shorthand. If a CLI app uses curlies for an input that comes back empty in a shorthand request, won't we do the same?

We could totally limit the surface area and check for `is_ui`, but I think it's really affecting all v3 integrations—not just UI.

xavdid:
Totally. I think this is a good way to do it, just that "changing default behavior" sets off warning bells. 😁 should be good though!

bcooksey:
I think this all checks out. I too was worried about default behavior change, but thinking a bit more on it, this is just being helpful. There is never a reason to send curlies, so replacing those with empty string by default now is a slight improvement. And then the option is there to strip the key out completely if the dev needs it.

bcooksey (approved):

PR #154, opened 2019-05-29
fix regression where env is ignored

xavdid:
Ok! So this fixes https://github.com/zapier/zapier/issues/26273. The root issue was that the early `require(appPath)`. Imagine you have the following:

```js
// app/index.js
module.exports = {
  // ... whatever
  url: `${process.env.BASE_URL}/oauth`,
}
```

when you `require('./app')`, `process.env.BASE_URL` is evaluated. The only problem is, the `env` being passed into lambda from the server isn't merged into `process.env` until [this line](https://github.com/zapier/zapier-platform-core/compare/fix-missing-env?expand=1#diff-b1d9d87090de25e8dcacf07d34e2e495R189), so `app.url` is `undefined/oauth`. When it's required again as part of `loadApp` further down, the result is cached. So even though `process.env.BASE_URL` exists, the url is wrong.

I read carefully through everything the http patch does and moving it further into that method should have no ill effects (since nothing between its original and new position can use the `http` module). All the tests are passing as well. I also tested locally and the current env pops up as expected.

I tried the following in `loadApp`, but weirdly, it didn't work. I had sunk enough time into this as is, so I didn't dig much further after confirming the fix itself.

```js
delete require.cache[require.resolve(appRawOrPath)]; // always get a fresh copy of the app
return resolve(require(appRawOrPath));
```

### DIFF
--- a/src/tools/create-lambda-handler.js
+++ b/src/tools/create-lambda-handler.js
@@ -130,14 +130,8 @@ const loadApp = (event, rpc, appRawOrPath) => {
   });
 };
 
-const createLambdaHandler = (appRawOrPath, { skipHttpPatch } = {}) => {
+const createLambdaHandler = appRawOrPath => {
   const handler = (event, context, callback) => {
-    // Adds logging for _all_ kinds of http(s) requests, no matter the library
-    if (!skipHttpPatch) {
-      const httpPatch = createHttpPatch(event);
-      httpPatch(require('http')); // 'https' uses 'http' under the hood
-    }
-
     // Wait for all async events to complete before callback returns.
     // This is not strictly necessary since this is the default now when
     // using the callback; just putting it here to be explicit.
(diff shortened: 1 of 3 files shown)


src/tools/create-lambda-handler.js
```diff
@@ -130,14 +130,8 @@ const loadApp = (event, rpc, appRawOrPath) => {
   });
 };
 
-const createLambdaHandler = (appRawOrPath, { skipHttpPatch } = {}) => {
+const createLambdaHandler = appRawOrPath => {
```
codebycaleb, line 133: I was going to say this looked dangerous because I thought any existing calls to `createLambdaHandler` that passed more than one argument would break, but it turns out JavaScript is pretty "meh" on enforcing function arity and would just drop it any way. 

It still might be good to fix any existing calls to this that have more than one argument provided (the one in `src/tools/create-app-tester.js` was the only one that I saw via a GitHub search but I didn't look very hard) in order to reduce future confusion (or errors if a new parameter is introduced later).

codebycaleb (approved):
I love when really complex problems (`env` is being ignored in a very specific set of circumstances) have simple solutions (move the `skipHttpPatch` handling). 😁 Thanks for nailing this one down!

stevelikesmusic (approved):
Nice find @xavdid! All looks good 👍

src/tools/create-lambda-handler.js
xavdid: yeah, js don't care. Typescript does though. In any case, good catch!

PR #157, opened 2019-06-07
Bump legacy-scripting-runner version to 3.1.0

bcooksey:
We need the bug fixes in 3.1.0 for converted Web Builder apps.


bcooksey:
cc @xavdid and @eliangcs I didn't realize this was still missing from core 8.2.1. I think what I'm gonna do is build a new boilerplate zip with 3.1.0 and push that up as the 8.2.1 zip

eliangcs (approved):

